PR #3, opened 2014-09-19
Fix non protected import of python dependences

nbessi:


pedrobaeza:
Can this module work without that dependencies?

nbessi:
No we may add a python-deps in manifest

pedrobaeza:
OK then. :+1:

yvaucher:
:+1:

PR #10, opened 2015-02-09
Show coordinates when overing

yvaucher:


base_geoengine/static/src/js/geoengine_view.js
```diff
@@ -578,9 +578,11 @@ openerp.base_geoengine = function(openerp) {
             var vl = new OpenLayers.Layer.Vector(self.name, {
 		styleMap: new OpenLayers.StyleMap({
 		    'default': new OpenLayers.Style({
-			fillColor: '#ee9900',
+			fillColor: '#ff0000',
+//			fillColor: '#ee9900',
```
yvaucher, line 582: To define if this color change is specific or not. We definitely don't want commented line there thought

pedrobaeza:
As this seems abandoned, I close. Feel free to reopen in case there's interest.

PR #13, opened 2015-02-09
[wip] Update the swisstopo WMTS urls

fredj:
Swisstopo has deprecated the `http://wmts[0-4].geo.admin.ch` range

### DIFF
--- a/base_geoengine/static/src/js/geoengine_view.js
+++ b/base_geoengine/static/src/js/geoengine_view.js
@@ -111,7 +111,7 @@ openerp.base_geoengine = function(openerp) {
                             name: l.name,
                             layer: l.swisstopo_type,
                             formatSuffix: 'jpeg',
-                            url: ['https://wmts0.geo.admin.ch/', 'https://wmts1.geo.admin.ch/', 'https://wmts2.geo.admin.ch/'],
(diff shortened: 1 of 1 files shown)


guewen:
:+1:

yvaucher:
@fredj Where can we find this info, as a reference?

I only found this doc and it seems it is outdated https://api3.geo.admin.ch/services/sdiservices.html#wmts

Nevertheless, you certainly know better than me. Then :+1:

fredj:
The doc is not outdated but it's a bit special for schweizmobil: they have an agreement with swisstopo to allow them to use the "private" `http://wmts[5-9].geo.admin.ch` range. This range bypasses the security proxy and the usage count system.

nbessi:
:+1:

nbessi:
@fredj is the branch still WIP?

yvaucher:
@nbessi yep it is still WIP in fact those url can have issues but are the public ones. The new ones are only for user with a contract with swisstopo as @fredj explained me.

fredj:
standby for now

fredj:
Not needed; `http://wmts[0-4].geo.admin.ch` are not deprecated, closing.

PR #14, opened 2015-02-12
Update OpenLayers to v2.13.1

fredj:
Update OpenLayers from 2.12 to 2.13.1.

See https://github.com/openlayers/openlayers/blob/master/notes/2.13.md


yvaucher:
:+1:

PR #16, opened 2015-02-12
7.0 migrate from 6.1

lmignon:
Hi,

I've adapted the code to work on the 7.0 branch.
- the map is rendered on the form view AND the geoengine view
- the readonly attribute is supported by the  geo_edit_map widget
- geonames_coder retrieves the username from the ir.config_parameter 'geoengine_geonames_username'  since geonames.org requires a username to access to the webservices.

Everything seems to work fine with OSM as raster layer type on res_partner. I've observed some troubles with google. (The spots appear shifted on the map. The google API is not loaded if you access to the form view before a first display of the geoengine view)

I have not fixed the  PEP8 errors to minimize changes and ease the review. (Will be done in another PR)

Regards

lmi


nbessi:
Hello thanks for the work. 
This review process is going to take a long time, I will give you feedback ASAP

nbessi:
@lmignon I should also provide you quickly the travis setup

yvaucher:
Thanks!

Ok for PEP8. I see that travis isn't even complaining certainly because it has not been configured.
This can be done in another PR

geoengine_geoname_geocoder/geonames_coder.py
```diff
@@ -79,6 +84,15 @@ def parse_code(code):
     def geocode_from_geonames(self, cursor, uid, ids, srid='900913', strict=True, context=None):
         context = context or {}
         base_url = u'http://ws.geonames.org/postalCodeSearch?'
+        config_parameter_obj = self.pool.get('ir.config_parameter')
```
yvaucher: Please use self.pool[]

lmignon:
@nbessi @yvaucher  Thank you for the future review. At this stage the code needs more cleaning and I only does the changes required to make the module working in 7.0. That's the first time I uses it and doing it working on 7.0 has been a good exercise to discover it.

lmignon:
git rebase

bwrsandman:
Hi @lmignon, do you think it's possible to get a non-minimized version of the js in geospatial/base_geoengine/static/src/js/lib/

lmignon:
@bwrsandman, js in the lib directory are external libraries used by the addons. Do we really need the non-minimized version of these libs?  @nbessi and @yvaucher  what do you think?

bwrsandman:
Last year when I was debugging it, it was a great pain to unminimize. Considering odoo minimizes when you don't have debug mode activated, it would be much better to have the code readable in the source.

lmignon:
@bwrsandman makes sense. I would prefer to make this change in an other PR.

geoengine_geoname_geocoder/geonames_coder.py
```diff
@@ -33,11 +33,16 @@
 
 logger = logging.getLogger('GeoNames address encoding')
 
+
 class ResPartner(osv.osv):
```
bwrsandman, line 37: Please use `orm.Model`

geoengine_geoname_geocoder/geonames_coder.py
```diff
@@ -33,11 +33,16 @@
```
bwrsandman, line 33: Since this file only defines ResPartner, could you rename it to res_partner.py

geoengine_geoname_geocoder/geonames_coder.py
```diff
@@ -33,11 +33,16 @@
 
 logger = logging.getLogger('GeoNames address encoding')
 
+
 class ResPartner(osv.osv):
     """Auto geo coding of addresses"""
     _name = "res.partner"
     _inherit = "res.partner"
```
bwrsandman, line 40: `_name` is redundant in this case, it would be cleaner without it

geoengine_geoname_geocoder/geonames_coder.py
bwrsandman, line 37: Also, please remove `ResPartner()` at the end of the file, both are deprecated forms

lmignon:
@bwrsandman @nbessi @yvaucher I had originally planned to make the migration to 7.0 in 2 steps. The first one, to make the code working in 7.0 (the current PR) and the second one to clean the code (pep8 + travis + replace deprecated import, etc....). If you prefer that the two steps are done in the same PR, you can label this PR with WIP. Whatever your decision, I hope to provide the second part of the changes early next week.

yvaucher:
@lmignon Go ahead for another PR for pep8 issues. BTW we will need to setup travis on this repo

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+
+install:
+  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
+  - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - travis_install_nightly
+  - pip install geojson Shapely
+  - sudo apt-get -q -y install postgis
+
+script:
+  - travis_run_flake8
```
bwrsandman: Calling `travis_run_flake8` is deprecated now that it's done in `travis_run_tests`

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+
+virtualenv:
+  system_site_packages: true
+
+install:
+  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
+  - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
+  - travis_install_nightly
+  - pip install geojson Shapely
+  - sudo apt-get -q -y install postgis
```
bwrsandman, line 19: Could these additional dependencies be mentioned in the README.md?

yvaucher:
[IMP] Add travis and coverage badges <- Great! Thanks for that

.travis.yml
yvaucher, line 19: https://github.com/OCA/geospatial/blob/8.0/__unported__/base_geoengine/README

For 7.0 I'm not sure what would be the best as modules don't support `README.rst`.
So It could wait for 8.0 version to merge `__openerp__.py` description and `README` into an `README.rst` file 

However, it is a good idea to give more visibility for those requirements in 7.0.

lmignon:
@bwrsandman done and you can see that only pep8 tests fails... :smile:

bwrsandman:
@lmignon nicely done. I think you may have missed my comments in  `geoengine_geoname_geocoder/geonames_coder.py`. I have a few issues with it that could be improved with the migration.

lmignon:
@bwrsandman yes I missed it. It's done now.

.travis.yml
yvaucher, line 19: (Done)

Thanks

README.md
```diff
@@ -40,3 +43,15 @@ Extending
...
+Requirements
+------------
+
+* System:
+    **PostGIS** http://postgis.refractions.net/
+
+* Python:
+    **Shapely** http://pypi.python.org/pypi/Shapely
+
+    **geojson** http://pypi.python.org/pypi/geojson
```
bwrsandman, line 57: This doesn't really format well in the richdiff, try adding a sub list:

``` markdown
* System:
    * **PostGIS** http://postgis.refractions.net/

* Python:
    * **Shapely** http://pypi.python.org/pypi/Shapely
    * **geojson** http://pypi.python.org/pypi/geojson
```

geoengine_geoname_geocoder/res_partner.py
```diff
@@ -33,11 +33,16 @@
 
 logger = logging.getLogger('GeoNames address encoding')
 
+
 class ResPartner(osv.osv):
```
bwrsandman, line 37: `osv.osv` -> `orm.Model`

geoengine_geoname_geocoder/res_partner.py
```diff
@@ -33,11 +33,16 @@
...
     """Auto geo coding of addresses"""
     _name = "res.partner"
     _inherit = "res.partner"
```
bwrsandman, line 40: `_name = "res.partner"` is redundant and can be removed

geoengine_geoname_geocoder/__openerp__.py
```diff
@@ -36,7 +41,7 @@
  'license': 'AGPL-3',
  'website': 'http://openerp.camptocamp.com',
  'depends': ['base', 'sale', 'geoengine_partner'],
- 'installable': False,
+ 'installable': True,
  'active': False,
  'icon': '/base_geoengine/static/src/images/map_icon.png'}
 # vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
bwrsandman, line 47: Would you mind removing the editor line?

bwrsandman:
Three more issues from both `geoengine_geoname_geocoder/res_company.py` and `geoengine_geoname_geocoder/res_partner.py`:
- `osv.osv` is deprecated and should be replaced with `orm.Model`
- `_name` is defined while there is a `_inherit` this is unnecessary, the module would be cleaner and more simple without it 
- At the end of the files the class is called, this is no longer necessary and should be removed (`ResPartner()`, `ResCompany()`)

lmignon:
@bwrsandman Thank you for the review but as explained here above, this PR is done to highlight the changes required to make the module working on 7.0. All others changes (pep8 , pylint, replace deprecated import, etc....) are done in an other PR #22.  The motivation behind these 2 PR is to ease the review since it's difficult to dissociate cosmetic changes from technical changes in a single PR. I wonder if it's a good idea to keep these two PRs since this introduces some confusion in the review process?!?

bwrsandman:
:+1: 

OK, sorry that I didn't see.

PR #39, opened 2015-06-23
fix imports to allow installation

ovnicraft:
Working to 7.0 geospatial i test it and need this fixes.

@nbessi please review

### DIFF
--- a/base_geoengine/__init__.py
+++ b/base_geoengine/__init__.py
@@ -19,17 +19,17 @@
 #
 ##############################################################################
 """The GeoEngine module"""
-from . import geo_model
-from . import geo_operators
-from . import geo_view
-from . import geo_helper
-from . import geo_ir
-from . import geo_db
-from . import geo_field
-from . import geo_point
-from . import geo_multipoint
-from . import geo_line
-from . import geo_multiline
-from . import geo_polygon
-from . import geo_multipolygon
-from . import company
+import geo_model
+import geo_operators
+import geo_view
+import geo_helper
+import geo_ir
+import geo_db
+import geo_field
+import geo_point
+import geo_multipoint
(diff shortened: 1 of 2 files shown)


base_geoengine/__init__.py
```diff
@@ -19,17 +19,17 @@
...
-from . import geo_db
-from . import geo_field
-from . import geo_point
-from . import geo_multipoint
-from . import geo_line
-from . import geo_multiline
-from . import geo_polygon
-from . import geo_multipolygon
-from . import company
+import geo_model
```
pedrobaeza, line 22: Why are you removing this relative import?

nbessi:
@ovnicraft that seems strange can you give me your version of Python?

lmignon:
@ovnicraft Thank you for your PR.  However I have no problem when installing this module on a fresh instance of Odoo 7.0.  (tests done now) Your problems seem related to how your environment is configured. The only circumstances in which I have observed problems with relative imports in Odoo is when I tried to run the tests directly from my IDE using the Unittest2 nor Nose runner. If this problem occurs in this case, be aware that it is not possible to run the tests without going through the Odoo runner. Indeed, the Odoo patch the PYTHONPATH at startup to insert the addons under the module 'openerp.addons'.

ovnicraft:
@nbessi 2.7.6 on osx
@lmignon i am working with virtualenv

nbessi:
@ovnicraft that is quite strange. if you add a `from __future__ import absolute_import` did it change something. As far as I remember there was some oddities with OSX python with this

ovnicraft:
@nbessi yes totally strange, btw i reproduce it in my vagrant machine (on OSX too).
Will test in my archlinux and give feedback.

PR #51, opened 2015-07-26
Update installation instructions

dreispt:
Now in proper place I hope.
I also added a note mentioning that OCB required no fixes to be applied.


lmignon:
:+1:  Thank you @dreispt

pedrobaeza:
:+1:

PR #60, opened 2015-11-18
[PORT] [9.0] base geoengine

yvaucher:


base_geoengine/__openerp__.py
```diff
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 {'name': 'Geospatial support for OpenERP',
- 'version': '8.0.0.2.0',
+ 'version': '9.0.0.2.0',
```
lmignon: we can start with version `9.0.1.0.0`

yvaucher:
For the big change (separating widgets from view file) some readable diffs without the move and the reindent:

http://pastebin.com/zs3k5SyM
http://pastebin.com/XLGY24bY

yvaucher:
@lmignon Here is a first usable version of geoengine using Openlayers 2

I didn't had time to take your in progress portage for Openlayers 3. But this should constitute as the base to take your work from 8.0 as I solved the issues of Odoo web framework changes.

lmignon:
@yvaucher Thank you for this work. No stress with Openlayers3 you already have done a big part of the job by restructuring and cleaning the js files. I have not had time to do a complete code review but a first overview already gives a very good impression. Should we still patch Odoo to be able to isntall the demo data?

yvaucher:
I used the same patch OCA/OCB#350

It seems to be still necessary looking at the odoo code:
https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L4229

yvaucher:
I tried without it and I get:

`ParseError: "Geometry SRID (0) does not match column SRID (900913)`

(IIRC it is the error for which we had to patch odoo)

base_geoengine/static/src/js/views/geoengine_view.js
```diff
@@ -0,0 +1,508 @@
...
+        // with no feature selected
+        this.selectFeatureControls = [
+            new OpenLayers.Control.SelectFeature(
+            {selectedFeatures:[]}, {hover: true, highlightOnly: true}),
+            new OpenLayers.Control.SelectFeature(
+                {selectedFeatures:[]}, {})
+        ];
+    },
+    limit: function() {
+        return false;
```
yvaucher: This return was already there in 8.0. Was it for debug ? Not sure we even need limit there. Thus I commented code for clarity and to remove Warning message in javascript console.

@lmignon any reason for this ? Should I remove the return false or the code below?
:smiling_imp: https://github.com/OCA/geospatial/commit/915925e206f81b1caf409323df6cd6b71be54278#diff-f6f3909c744c4f386dfa5d5cc1d99cccR171

base_geoengine/static/src/js/views/geoengine_view.js
```diff
@@ -0,0 +1,508 @@
...
+        // use {selectedFeatures:[]} as a hack to simulate a vector layer
+        // with no feature selected
+        this.selectFeatureControls = [
+            new OpenLayers.Control.SelectFeature(
+            {selectedFeatures:[]}, {hover: true, highlightOnly: true}),
+            new OpenLayers.Control.SelectFeature(
+                {selectedFeatures:[]}, {})
+        ];
+    },
+    limit: function() {
```
lmignon: IMO you could remove limit and offset from the code... These are never used and I should have already removed this code a long time ago...

lmignon:
> `ParseError: "Geometry SRID (0) does not match column SRID (900913)`
> @yvaucher Yes it is the error for which we had to patch odoo.

yvaucher:
I just removed the limit and offset methods.

lmignon:
@yvaucher Thank you. I hope to have some time to review your PR in the next days. Sorry for the delay...

yvaucher:
@lmignon No worries, when you do just ping me and I'll fix it.

yvaucher:
@fredj Could you have a look at the patch on mapfish ? Maybe it's a misuse of the lib, otherwise I'll report it on mapfish project

base_geoengine/static/src/js/views/geoengine_widgets.js
```diff
@@ -0,0 +1,420 @@
...
+                },
+                scope: this
+            }
+        });
+        var rl = this.createBackgroundLayers([field_infos.edit_raster]);
+        rl.isBaseLayer = true;
+        return [rl, vl];
+    },
+
+    add_tab_listener: function() {
```
yvaucher, line 72: This tab listener doesn't work anymore.
If a map widget in on second tab it won't be displayed

To fix!

yvaucher:
And now we are good.

base_geoengine/static/src/js/views/geoengine_widgets.js
yvaucher, line 72: Fixed

yvaucher:
@lmignon Any idea why travis fails on DOCS ?

Added some changes:
- Fixed issue with geo widget in tabs.
- Use of translated values in geoengine view instead of raw values (field name and selection values)

lmignon:
@yvaucher travis seems to fail when applaying the patch https://github.com/yvaucher/geospatial/blob/9.0-port-base_geoengine/.travis.yml#L34
https://travis-ci.org/OCA/geospatial/jobs/104121499#L296

yvaucher:
@lmignon I'll have a look at that this week

yvaucher:
@lmignon This should work, I had to make another patch for official odoo branch 9.0 as the code base changed since 8.0, plus now OCB is patched.
https://github.com/odoo/odoo/pull/10639

lmignon:
@yvaucher Thank you... I wonder why this patch is not yet merged.... :sleepy:

yvaucher:
Hum there is an error when trying to create an object with a geometry.

DataError: Geometry SRID (0) does not match column SRID (900913)

It works though with a write. (creating the object without the geom and then edit it to add the geom)

lmignon:
@yvaucher I encountered the same kind of error when the column._synbol-c was not used by odoo to generate the SQL statement. May be a case not covered by the patch...

yvaucher:
@lmignon That's what I figured out thus we are pretty ready here with some added features.

yvaucher:
@lmignon Do you mind if we merge this PR ? I'm starting to have other incoming features

lmignon:
@yvaucher  the code looks good to me but I've no time to install the code to make a  functional test. Can you remove the patch applied on odoo in .Travis.y ml? The fix is merged d82e489add442f0bef52b58620c8d7988919872d

yvaucher:
@lmignon Removed patch and fixed the doc build

lmignon:
@yvaucher Thank you for all this work. Travis is now green! 
You have my :+1:  for the merge.

yvaucher:
@gurneyalex Thanks, runbot is working now !

@lmignon Maybe you can have a look at it via runbot.

gurneyalex:
:+1:

lmignon:
@yvaucher Tested on runbot. Works fine. Great job! Thank you @gurneyalex for runbot.

Yenthe666:
@yvaucher when I install this on a V9 enterprise I will get the following issue:

```
  File "/odoo9ent/odoo9ent-server/addons/base_geoengine/geo_helper/geo_convertion_helper.py", line 46, in value_to_shape
    shape_to_return = wkt.loads(value)
ParseError: "global name 'wkt' is not defined" while parsing /odoo9ent/odoo9ent-server/addons/base_geoengine_demo/data/retail_machine_geom.xml:4, near
<record model="geoengine.demo.automatic.retailing.machine" id="machine_1">
         <field name="name">1</field>
         <field name="total_sales">1057</field>
         <field name="money_level">low</field>
         <field name="the_point">POINT(746676.106813609 5865349.7175855)</field>
         <field name="state">ok</field>
      </record>
```

lmignon:
@Yenthe666 

```
pip install geojson Shapely
```

Yenthe666:
@lmignon it is installed. It all seems to work fine on a regular V9 but not on a V9 Enterprise I think.

lmignon:
@Yenthe666 I can't figure out what's could be cause this kind of issue since the only difference between Entreprise and Community in the case of the geoengine addons is the addon 'web'

Yenthe666:
@lmignon a reboot seems to have fixed everything. I will try a clean install again tommorow to double check this!
It works fine on V9 (community) for sure

lmignon:
@Yenthe666 Thank you for the report. You are welcome to give your :+1: to @yvaucher. New reviewers are welcome :smirk:

Yenthe666:
@lmignon looks like it was an issue somewhere on the VM since a clean install seemed to have worked fine! I do find some jQuery / JS issues if I change some map settings. Is there any tutorial / guide on what is allowed or what is not? It is really easy to break the whole maps.

lmignon:
@Yenthe666 Thank you for the feedback. Can you describe more precisely the kind of issues your encounters and the steps required to reproduce these issues?

yvaucher:
@Yenthe666 Thanks for the feed back, it would be great if you could give us more details. For now there are 3 bugs I know in this version which might be related to the edition tools of the widget. However, reloading the page solve the issue in all case.

Issue 1: (I think it was already there in 8.0)
The polygon widget when creating a shape show the current shape and stay in creating mode which make the previously already created shape plus the shape being created at the same time. Finishing a new shape erase the former shape.

Issue 2: (Also have seen it occasionally since 6.1)
A point is sometimes displayed twice in a widget.

Issue 3:
With a map in a second tab, it happen some times in create mode that the map it displayed twice. This might be the editable layer and the read-only layer in 2 separate maps. This needs further investigations.

And more with things such as google layer which doesn't work (but also in 7.0 and 8.0)

Have you encountered other issues ?

Still, I would prefer to merge this non perfect but working version to be able to fix case by case the other issues.
It would also help on starting the migration other modules of this repository.

Yenthe666:
@yvaucher none of those 3 issues I think.
For example, the original ZIP OSM raster layer:
image
Changing it to Google maps:
image
Result on the view:
image

It seems that the Google views are not rendered? Or is this expected behaviour?

Another example. Original retail machine mapbox:
image

Changing the raster layer type to 'Distant WMS':
image

Next when you go back to the view everything is broken:
image

It breaks all views and you'll even need to refresh pages to fix this.

Changing retail machine mapbox to the raster layer type swisstopo:
image

The view:
image

Does this help you in any way? There are quite a lot of these issues but it might be me doing things wrong I guess.

yvaucher:
@Yenthe666 

For google layer this is a known bug in all versions #63 
We will focus on making OSM map work first. google layer is secondary.

For swisstopo these tiles are restricted to switzerland coutry area thus tiles are not available for most part of the world.
I have extracted this in 2 other PR: #72 #74 
Plus in #72 it is now possible to define a restricted extent.

lmignon:
@yvaucher I propose to remove google from the list of available layer at this stage.  Moreover, google is not supported with ol3. IMO, we should completely drop google from the code.

Yenthe666:
@yvaucher and @lmignon I'd really love the Google functionality, it was looking really interesting to me. Its a major plus since most people in the world use Google Maps.

yvaucher:
@lmignon @Yenthe666 

On one hand google layer is not working right now, thus in a first move I'll change hide the value from selection field.

On the other hand google layers might be helpful for someone.
And removing it would break other PR waiting on this one.
Plus it seems possible to use it in OL3:
http://openlayers.org/en/v3.0.0/examples/google-map.html

So for now I will only comment out the value in the selection adding a FIXME.
Maybe we would like to extract it in an other module but this kind of refactor is out of scope of this PR.

lmignon:
@yvaucher Ok for removing the value from the selection field.

Yenthe666:
:+1: let us go ahead, merge this so we can get things going :smile: and thanks for all the work guys!

yvaucher:
:tada:

guewen:
Excited

PR #65, opened 2015-11-30
[7.0] Fix quantil repartition when there are less item than classes

### LINKED ISSUE #62: [8.0] [9.0] geoengine_demo: TypeError: bin.upperBound is undefined on some search
In demo data, when doing a search by code on retail machines it can raise the following javascript error:

Tested on port of 9.0 and on 8.0

To reproduce:
Open retails machine geoengine view.

Do a search on `3` -> 9 results it works
Do a search on `4` -> 4 results it fails
Do a search on `7` -> 4 results it fails
Remove code 37 do a search on `7`-> 3 results it fails
Remove code 27 do a search on `7` -> 2 results it passes

```
TypeError: bin.upperBound is undefined

http://localhost:8501/base_geoengine/static/src/js/mapfish/GeoStat.js:351
Traceback:
mapfish.GeoStat.Distribution<.defaultLabelGenerator@http://localhost:8501/base_geoengine/static/src/js/mapfish/GeoStat.js:351:16
mapfish.GeoStat.Distribution<.classifyWithBounds@http://localhost:8501/base_geoengine/static/src/js/mapfish/GeoStat.js:383:29
mapfish.GeoStat.Distribution<.classifyByQuantils@http://localhost:8501/base_geoengine/static/src/js/mapfish/GeoStat.js:415:16
mapfish.GeoStat.Distribution<.classify@http://localhost:8501/base_geoengine/static/src/js/mapfish/GeoStat.js:452:30
mapfish.GeoStat.Choropleth<.setClassification@http://localhost:8501/base_geoengine/static/src/js/mapfish/GeoStat/Choropleth.js:141:31
GeoengineView<.createVectorLayer@http://localhost:8501/base_geoengine/static/src/js/views/geoengine_view.js:230:13
GeoengineView<.createVectorLayers/<@http://localhost:8501/base_geoengine/static/src/js/views/geoengine_view.js:299:22
_.forEach@http://localhost:8501/web/static/lib/underscore/underscore.js:145:9
```

Tracing it a bit it seams it fails on `geostat.setClassification()` when it calls a method `classify` it passes `this.numClasses` as `nbBins` with a value of 5 for 4 items. Then it does a strange
 `var binSize = Math.round(this.values.length / nbBins);`

Which will round to 5, thus looping to compute bounds, it tries to take in account the bound of an non existing 5th object, and fail.

yvaucher:
Fixes #62

### DIFF
--- a/base_geoengine/static/src/js/mapfish/GeoStat.js
+++ b/base_geoengine/static/src/js/mapfish/GeoStat.js
@@ -399,14 +399,15 @@ mapfish.GeoStat.Distribution = OpenLayers.Class({
     classifyByQuantils: function(nbBins) {
         var values = this.values;
         values.sort(function(a,b) {return a-b;});
+        // We don't want more classes that elements to classify
+        var maxNbBins = Math.min(nbBins, values.length - 1);
         var binSize = Math.round(this.values.length / nbBins);
 
         var bounds = [];
         var binLastValPos = (binSize == 0) ? 0 : binSize;
-
         if (values.length > 0) {
             bounds[0] = values[0];
-            for (i = 1; i < nbBins; i++) {
+            for (i = 1; i < maxNbBins; i++) {
                 bounds[i] = values[binLastValPos];
                 binLastValPos += binSize;
             }
--- a/base_geoengine/static/src/js/mapfish/GeoStat.js
+++ b/base_geoengine/static/src/js/mapfish/GeoStat.js
@@ -401,7 +401,7 @@ mapfish.GeoStat.Distribution = OpenLayers.Class({
         values.sort(function(a,b) {return a-b;});
         // We don't want more classes that elements to classify
         var maxNbBins = Math.min(nbBins, values.length - 1);
-        var binSize = Math.round(this.values.length / nbBins);
+        var binSize = Math.round(this.values.length / maxNbBins);
 
         var bounds = [];
         var binLastValPos = (binSize == 0) ? 0 : binSize;


base_geoengine/static/src/js/mapfish/GeoStat.js
```diff
@@ -399,14 +399,15 @@ mapfish.GeoStat.Distribution = OpenLayers.Class({
     classifyByQuantils: function(nbBins) {
         var values = this.values;
         values.sort(function(a,b) {return a-b;});
-        var binSize = Math.round(this.values.length / nbBins);
+        // We don't want more classes that elements to classify
+        var maxNbBins = Math.min(nbBins, values.length - 1);
```
pedrobaeza, line 403: Please avoid uppercase in local variable names

pedrobaeza, line 403: Why not using native `min` method?

lmignon, line 403: @pedrobaeza uppercase in local variables names is by convention in this file... IMO it's better to preserve consistency.

pedrobaeza, line 403: OK

lmignon, line 403: @pedrobaeza `min` method is not native.
For ex in Chrome

``` javascript
min
VM132:2 Uncaught ReferenceError: min is not defined(…)
```

lmignon:
:+1: Thank you for the fix and the port to others branches! (Code review, no test)

base_geoengine/static/src/js/mapfish/GeoStat.js
pedrobaeza, line 403: Sorry, I got confused with Python. Very late review...

pedrobaeza:
:+1:

base_geoengine/static/src/js/mapfish/GeoStat.js
lmignon, line 403: > Sorry, I got confused with Python. Very late review...

That's why multi reviewers is always a good practice. We are all humans.. :smirk:

pedrobaeza, line 403: Indeed!

yvaucher, line 403: You scared me with so much talk for one line, I was wondering what I did so wrong. :sweat_smile:  

Thanks for the review anyway.

fredj:
+1

PR #72, opened 2016-02-12
[9.0]	Refactor WMTS

yvaucher:
Depends on 
- [x] #60 

Allow to configure WMTS by parameters.

Extract swisstopo to move it in a new module.

### DIFF
--- a/base_geoengine/geo_view/geo_raster_layer.py
+++ b/base_geoengine/geo_view/geo_raster_layer.py
@@ -1,24 +1,22 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Author: Nicolas Bessi
-#    Copyright 2011-2012 Camptocamp SA
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


lmignon:
@yvaucher Can you rebase your branch?

base_geoengine/geo_view/geo_raster_layer.py
```diff
@@ -28,34 +42,103 @@ class GeoRasterLayer(models.Model):
...
+    @api.depends('raster_type')
+    def _get_is_wmts(self):
+        self.is_wmts = self.raster_type in ('swisstopo', 'wmts',
+                'map.lausanne.ch')
+
+    @api.one
+    @api.onchange('raster_type')
+    def onchange_set_wmts_options(self):
+        # XXX move in swisstopo module
+        if self.raster_type == 'swisstopo':
```
yvaucher: To clean up

base_geoengine/geo_model.py
```diff
@@ -178,7 +178,8 @@ def get_edit_info_for_geo_column(self, cursor, uid, column, context=None):
         res['edit_raster'] = raster_obj.read(
             cursor, uid, raster_id[0], context=context)
         res['geo_type'] = field.geo_type
-        res['srid'] = field.srid
+        # get srid from _columns as srid is wrong in _fields
```
lmignon: I think that this code will not work if your field is a related. In this case, `_columns .get(column)` is `None` since a related field in the new API has no column associated to the field.

yvaucher: Ok, on the other hand, `srid` has always the default value 900913 while `_srid` is set to the right one defined on the field.

There might be something missing in `base_geoengine/geo_field.py`

I'll have to check deeper in the code. Thus I put this as WIP

yvaucher: @lmignon I extracted it in #79 as it's not part of the refactoring

yvaucher:
Rebased

yvaucher:
Runbot rebuilt, don't know why it failed

lmignon:
:+1:  Nevertheless I wonder if would be more modular to extract each layer type in independent module..

yvaucher:
Let's proceed with this one

PR #73, opened 2016-02-12
[9.0] base_geoengine - restricted extents

yvaucher:
Depends on
- [x] #60 
- [x] #72

To work properly with projection it needs #79 

Add restricted extent option on view

### DIFF
--- a/base_geoengine/geo_model.py
+++ b/base_geoengine/geo_model.py
@@ -1,24 +1,7 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Author: Nicolas Bessi
-#    Copyright 2011-2012 Camptocamp SA
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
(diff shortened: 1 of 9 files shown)


lmignon:
@yvaucher a rebase is needed.

yvaucher:
Rebased

base_geoengine/geo_model.py
```diff
@@ -178,7 +163,10 @@ def get_edit_info_for_geo_column(self, cursor, uid, column, context=None):
         res['edit_raster'] = raster_obj.read(
             cursor, uid, raster_id[0], context=context)
         res['geo_type'] = field.geo_type
-        res['srid'] = field.srid
+        # get srid from _columns as srid is wrong in _fields
+        res['srid'] = self._columns.get(column)._srid
```
lmignon: If the field is not stored, `self._column.get(column)` will return None

lmignon: @yvaucher see https://travis-ci.org/OCA/geospatial/jobs/120637898#L481 That's why you can't get the srid from the column...

yvaucher: Better to discuss this here #79

lmignon:
@yvaucher Usefull new functionnality :+1:  Code review + Functional tests

guewen:
:+1:

PR #78, opened 2016-02-17
[9.0] fix duplicate map in form view tabs

yvaucher:
When having a map in a form, inside a second tab it happens that the map of other object where displayed when creating a new object.

This was due to some issue with the tab listener to trigger the rendering of the map.

This PR fix this.

### DIFF
--- a/base_geoengine/static/src/js/views/geoengine_widgets.js
+++ b/base_geoengine/static/src/js/views/geoengine_widgets.js
@@ -70,15 +70,14 @@ var FieldGeoEngineEditMap = common.AbstractField.extend(geoengine_common.Geoengi
     },
 
     add_tab_listener: function() {
-        var self = this;
-        $('a[data-toggle="tab"]').on('shown.bs.tab', function(e) {
-            var geo_tab_id = self.$el[0].id;
-            var active_tab = $(e.target.hash);
-            if (active_tab.find('#' + geo_tab_id).length == 1) {
-                  self.render_map();
-                  return;
-            }
-        });
+        tab_href = this.$el.closest('div[role="tabpanel"]')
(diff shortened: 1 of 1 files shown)


base_geoengine/static/src/js/views/geoengine_widgets.js
```diff
@@ -70,15 +78,14 @@ var FieldGeoEngineEditMap = common.AbstractField.extend(geoengine_common.Geoengi
...
-        var self = this;
-        $('a[data-toggle="tab"]').on('shown.bs.tab', function(e) {
-            var geo_tab_id = self.$el[0].id;
-            var active_tab = $(e.target.hash);
-            if (active_tab.find('#' + geo_tab_id).length == 1) {
-                  self.render_map();
-                  return;
-            }
-        });
+        tab_href = $('#' + this.name).closest('div[role="tabpanel"]')
```
lmignon: why not:

```
this.$el.closest('div[role="tabpanel"]').on
```

IMO, it's a little bit safer since name is not unique

yvaucher: Indeed I should better use `this.$el`

However, I can't attach the trigger directly to the pane content. The event "shown.bs.tab" happen on the tab label button.

yvaucher:
Here we still have issues, especially when oppening a popup containing a map widget in a tab.

yvaucher:
Now the widget should work correctly in tabs and in popups

guewen:
:+1: (code review)

lmignon:
:+1: ( Code review)

yvaucher:
Travis is complaining only on sphinx doc, I'm merging

PR #79, opened 2016-02-18
[9.0] Fix srid value read from column

yvaucher:
self._fields.get(column).srid returns always `900913` even when a different srid is set on the field.

### DIFF
--- a/base_geoengine/fields.py
+++ b/base_geoengine/fields.py
@@ -39,9 +39,12 @@ class GeoField(Field):
     """
 
     geo_type = None
-    dim = 2
-    srid = 900913
-    gist_index = True
+
+    _slots = {
+        'dim': 2,
+        'srid': 900913,
+        'gist_index': True,
+    }
 
     def convert_to_read(self, value, use_name_get=True):
         res = convert.value_to_shape(value)


lmignon:
I see an other inconsistency in the code... 
 https://github.com/yvaucher/geospatial/blob/9.0-fix-srid/base_geoengine/geo_model.py#L88 (col.srid <> col._srid)

lmignon:
@yvaucher  I wonder if the bug doesn't comes from the fact that `srid` is not declared in the `_slots` on the GeoField.  Can you try:

``` python
class GeoField(Field):
    """ The field descriptor contains the field definition common to all
    specialized fields for geolocalization. Subclasses must define a type
    and a geo_type. The type is the name of the corresponding column type,
    the geo_type is the name of the corresponding type in the GIS system.
    """

    geo_type = None
    _slots = {
        'dim': 2,
        'srid': 900913,
        'gist_index': True
    }

    ....
```

yvaucher:
@lmignon Thanks for the hint I'll try this, maybe later this week

yvaucher:
@lmignon I fixed it like you suggested and it seems to work perfectly.

lmignon:
@yvaucher great! ! Difficult to test but I trust you and according to the code in Odoo this is the way to declare properties of fields. I don't know why travis fails but that seems unrelated to your change. :+1:

yvaucher:
Indeed travis error is unrelated, it fails the same way in other PRs

yvaucher:
The sphinx error is due to https://github.com/OCA/odoo-sphinx-autodoc/issues/3

Olivier-LAURENT:
:+1:  Code review only. Thanks for the fix.

PR #87, opened 2016-04-20
Fix _setup_fields signature due to 87d160a7423243ecd760fefe5ba22fb85c…

yvaucher:
…460cb6

Otherwise partial value is interpreted as context and it will fails on
Odoo loading

See the changes in https://github.com/OCA/OCB/commit/87d160a7423243ecd760fefe5ba22fb85c460cb6#diff-6a59d8ada6e3d58bf74e9fc0691b50ac


pedrobaeza:
:+1:, but you must rebased due to the version number

yvaucher:
Rebased already

lmignon:
:+1:

guewen:
:+1:

PR #88, opened 2016-04-22
[9.0] Fix offset issues on the map by reseting offset when scroll bar is used

yvaucher:
This force recomputation of offsets after each time the scrollbar is used.


yvaucher:
Merging as it is an important fix and few line of code.

PR #89, opened 2016-04-22
[9.0] Avoid zoom on set_value + define default zoom

yvaucher:
This is a good improvement to geometry edition, it will avoid you several zoom outs you would have to do to restore the right zoom level you need to see everything you need.

### DIFF
--- a/base_geoengine/static/src/js/views/geoengine_widgets.js
+++ b/base_geoengine/static/src/js/views/geoengine_widgets.js
@@ -109,24 +109,26 @@ var FieldGeoEngineEditMap = common.AbstractField.extend(geoengine_common.Geoengi
         });
     },
 
-    set_value: function(value) {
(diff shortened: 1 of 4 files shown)


guewen:
:+1: (code review)

lmignon:
:+1: (code review)

base_geoengine/static/src/js/views/geoengine_widgets.js
```diff
@@ -109,24 +110,29 @@ var FieldGeoEngineEditMap = common.AbstractField.extend(geoengine_common.Geoengi
         });
     },
 
-    set_value: function(value) {
+    set_value: function(value, zoom=true) {
```
yvaucher: Not possible to pass kwargs. To replace by `options` dictionary

Issue #36, opened 2015-06-04
"Localize addresses from GeoNames" menu is broken


nbessi:

alexalv:
Facing the same issue in v8
![screenshot](https://cloud.githubusercontent.com/assets/450994/14285917/ea39c39e-fb55-11e5-9184-4504591c9bef.png)

alexalv:
I made 
`parent="base_geoengine.geoengine_base_view_menu"`
 in menuitem and added
`<field name="target">new</field>`
 in action, that seem to solve this problem in v8. Result can be seen on the screenshot below.
Should I make a pull-request or this is not a correct way to solve this issue?
![screenshot](https://cloud.githubusercontent.com/assets/450994/14287460/664aa1e2-fb5b-11e5-9249-416c19c986c4.png)

yvaucher:
Hello @alexalv I agree on the need the change the parent of the menu item.

And also as it is a wizard we want it as "new" in a popup.

A PR would be great. And feel free to fix that typo too: "Geolocalize **you** addresses" -> your
Plus I see a boolean field with no label that miss a `<group>` around it.

And the buttons which could be in a `<footer>`.

PR #91, opened 2016-04-26
[8.0] geoengine_geoname_geocoder: new parent for wizard menu, buttons in footer

alexalv:
Fixes issue #36 
Changed 'parent' for menuitem which is calling partner selection wizard in geoengine_geoname_geocoder module. The same wizard is now called with target set to 'new', boolean 'encode_all' field is now in 'group' tags on wizard form, wizard buttons are now inside 'footer' tag.
Result

### DIFF
--- a/geoengine_geoname_geocoder/wizard/bulk_encode_view.xml
+++ b/geoengine_geoname_geocoder/wizard/bulk_encode_view.xml
@@ -15,8 +15,9 @@
                     <field name="add_to_encode" colspan="4" attrs="{'invisible': [('encode_all', '=', True)]}"/>
                     <newline/>
                     <footer>
-                        <button icon="gtk-cancel" special="cancel" string="Cancel"/>
-                        <button icon="gtk-ok" name="encode" string="Encode addresses" type="object"/>
+                        <button class="oe_hightlight" name="encode" string="Encode addresses" type="object"/>
+                        or
+                        <button class="oe_link" special="cancel" string="Cancel"/>
                     </footer>
                 </form>
             </field>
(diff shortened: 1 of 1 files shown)


yvaucher:
Thanks for the fix. Could you also:
- add `class="oe_hightlight"` on the button "Encode addresses"
- add `class="oe_link"` on the cancel button
- place cancel on the right as in all other wizards
- add an "or" between the 2 buttons

e.g.: https://github.com/OCA/OCB/blob/8.0/addons/stock/wizard/make_procurement_view.xml#L29-L31

Otherwise tested on runbot the wizard opens correctly.

alexalv:
@yvaucher thank you for the fast reply. Done.
Result

geoengine_geoname_geocoder/wizard/bulk_encode_view.xml
```diff
@@ -9,21 +9,24 @@
...
+                    </group>
                     <field name="add_to_encode" colspan="4" attrs="{'invisible': [('encode_all', '=', True)]}"/>
                     <newline/>
-                    <group colspan="4" col="6">
-                        <label string=" " colspan="2"/>
-                        <button icon="gtk-cancel" special="cancel" string="Cancel"/>
-                        <button icon="gtk-ok" name="encode" string="Encode addresses" type="object"/>
-                    </group>
+                    <footer>
+                        <button class="oe_hightlight" name="encode" string="Encode addresses" type="object"/>
```
yvaucher: Sorry there was a typo in my previous comment. It should be `oe_highlight` there is an extra `t`

alexalv:
@yvaucher I completely missed it too. Now fixed. 
Result

yvaucher:
:+1: Thanks

yvaucher:
Okay let's merge this it seems that we won't get a second review here.

PR #98, opened 2016-07-13
[9.0] Fix z-index of Openlayers map container to ensure it doesn't hide sea…

yvaucher:
…rch drop down list

Before:

bug-search-z-index

After:

fix-z-index


pedrobaeza:
:+1:

PR #104, opened 2016-09-01
[8.0] Backport Add geo_contains operator to geo_search method.

yvaucher:
geo_contains calls St_Contains SQL search.

Plus fix search to set SRID to avoid mixin srid issue.


guewen:
:+1:

PR #105, opened 2016-09-05
[8.0] Map selection

yvaucher:
This Pull request adds the possibility to select multiple feature by 2 ways:
1. Using the new selection tool to draw a selection rectangle
2. Pressing on shift key to add other features

Once more than one selected feature is selected a button is shown to allow you to filter by the selection.
This make a selection by ids of the selected features. And thus you will be able to switch to list view. For example to launch a wizard on all the matches. Or you can switch to form view to see details of each selected entries.

geoengine_selection_v8
geoengine_selection_in_list_v8

TODO: extend icon set to add a better selection icon.


guewen:
:+1:

feketemihai:
:+1:  Nice improvement, thanks for it...

pedrobaeza:
Please implement a better icon or include it in the Known issues section

pedrobaeza:
Why the word "x features selected" instead of "x records selected"?

yvaucher:
@pedrobaeza feature is the generic word for a geometry on the map as you could select a point, a line, a multi-line, a polygon or a multi polygon. A future improvement could be to put the name of the object to which the displayed geo field belongs.

lmignon:
@YannickB Nice improvement! It took me time to find out how to perform the selection.. It would be nice to describe how to use this new feature into the documentation....

pedrobaeza:
@yvaucher thanks for the explanation. The improvement can be also annotated in the "Known issues / Roadmap" section

yvaucher:
@lmignon wrong Yannick :grimacing: 

The PR for documentation must be done on 9.0 then. I can't do it in this PR. If you are talking about http://oca.github.io/geospatial/index.html

Which of the 2 type of selection wasn't well described in this PR summary to know what to write in readme in more details ?

lmignon:
@yvaucher No tooltip on the new tool into the toolbox. If the functionality is not described somewhere, the end user has no chance to discover it. :smirk: 
Nevertheless :+1:

yvaucher:
@lmignon okay tooltip plus correct icon should help. Will see if I have the time to do this later this week.

I was also thinking of a text to show somewhere to explain the SHIFT + click only visible with the 2 first tools (not usable with measure tools)

YannickB:
@yvaucher @lmignon I was wondering why I got a mention in geospatial module XD, now I understand. Good luck guys.

lmignon:
@YannickB Sorry my mistake

yvaucher:
Labelled as WIP as there is still an error on partner view when doing a selection + I have to had tool tip + better icon.

yvaucher:
Fixed the error when we try to select object without geometry (which were added to the map)

And added a more decent icon.

lmignon:
For the record, @yvaucher I've had the following error on runbot

```
Uncaught TypeError: Cannot read property 'toFixed' of undefined
```

image

The error has occurred when I've clicked on `the filter selection` button. It was may be a problem with the browser cache since this error no more occurs after emptying the browser cache.

yvaucher:
@lmignon I played a bit with it on runbot and can't reproduce.

It could be several issue as `toFixed` is called when setting style on svg nodes, and in mesure tool or could be in GeoStat.js.

Anyone else got the same issue ?

lmignon:
@yvaucher a browser cache issue :smirk:

PR #106, opened 2016-09-10
[FIX] from_latlon method srdi was incorrect

alan196:
This PR is to fix an issue that don't allow the method from_latlon to work.


alan196:
@yvaucher

yvaucher:
:+1: LGTM

PR #110, opened 2016-10-04
9.0 selection filter

yvaucher:
This is a port of the selection filter made for geoengine in odoo 8.0 on odoo 9.0 with OL3
- First commits are cherry-picked from https://github.com/OCA/geospatial/pull/105
- Instead of having a button in tool box, dragbox is usable by pressing the SHIFT key.
- It supports multi vector layer selection.

### DIFF
--- a/base_geoengine/static/src/js/views/geoengine_view.js
+++ b/base_geoengine/static/src/js/views/geoengine_view.js
@@ -88,7 +88,7 @@ var formatFeatureHTML = function(a, fields) {
  */
 var formatFeatureListHTML = function(features) {
     var str = [];
-    str.push(features.length + ' features selected');
+    str.push(features.getLength() + ' features selected');
     return str.join('<br />');
 };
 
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Can you squash a bit the commits?

base_geoengine/static/src/js/views/geoengine_view.js
```diff
@@ -524,6 +573,18 @@ var GeoengineView = View.extend(geoengine_common.GeoengineMixin, {
             });
             var layerSwitcher = new ol.control.LayerSwitcher({});
             map.addControl(layerSwitcher);
+            /*
```
pedrobaeza: Remove commented code

yvaucher:
@pedrobaeza Still work in progress and no I can't squash commits, most of the commits are cherry-pick from https://github.com/OCA/geospatial/pull/105

yvaucher:
@pedrobaeza BTW target branch is 9.0-ol3

pedrobaeza:
Yeah, I see, it would be good at the end of the process then

yvaucher:
@pedrobaeza still I wanted to keep track of commits done in 8.0 inside 9.0 to show that we are not missing those commits. Here you have few copied commits plus few new commits to port it.

pedrobaeza:
OK, it's not a heavy requirement, but you can see that a not-squashing operation in 8.0 makes now to appear a lot of useless commits. Let's try to make it better in the new 9.0 commits!!

PR #115, opened 2016-10-04
Fix map size error when open record in a fullscreen map.

cyrilgdn:
Bug was: Open map view full screen, open record form view (by clicking on a point), and switch to map view again => JS error because the map object size is undefined


yvaucher (approved):
Tested on runbot

PR #121, opened 2016-10-09
[9.0] ol3

yvaucher:
Portage of Geoengine to Openlayers 3

**Changes:**
- _Now based on Openlayers 3.18.2_
- Dropped mesurement tools (instead there is now a scale)
- change of cholopleth to chroma.js for quantile colors
- Forward ported selection tool to create filters from the map from 8.0 (Controls have been simplified, SHIFT key allows to use selection box)
- Dropped Google and Mapbox raster layer (both were deprecated in OL3)
- Adds a full screen mode on geoengine view
- Adds a example of onchange with geom in demo module


lmignon:
@yvaucher Is this PR ready for review or still a WIP?

yvaucher:
@lmignon This is a working version. It just deserve a major version change in `__openerp__.py`

But this could also break other modules or projects. Thus we might just want to keep the `9.0-ol3` branch on its own side to side with `9.0`, or at least make a tag to let people download the old version easily. Nevertheless, for 10 we will directly start on top of this even if it hasn't been merged in 9.0 branch.

Anyway this work deserve a review so this is the simplest way to do it in here.

yvaucher:
Tag created there:
https://github.com/OCA/geospatial/tree/9.0-ol2

yvaucher:
Closing, this branch stays available as a separate branch though

dek-odoo:
:+1

PR #123, opened 2016-10-23
10.0 base geoengine

yvaucher:
This is based on port to ol3 on v 9.0 #121 

Thus what changes for v 10.0 are in the last commits.

Mostly in https://github.com/OCA/geospatial/pull/123/commits/36cda2d


base_geoengine/fields.py
```diff
@@ -25,27 +28,140 @@ class GeoField(Field):
 
     geo_type = None
```
lmignon, line 29: IMO `geo_type` is no more required. You can directly use _type. Therefore the override of `ir.model.fields` will be no more needed

yvaucher: `geo_type` includes the postgis type for the field when using "AddGeometryColumn".

I don't think this value is suited to `type` which is kind of odoo name for a type of field.

This also allow to use type to check if it has a prefix in 
https://github.com/OCA/geospatial/pull/123/files#diff-758bcd153189712715dfd56ddc0c293dR42

ok an `isinstance()` would do the trick but this makes the type more readable, ie. `type='geo_line'` has more meaning than `type='linestring'`

yvaucher: I think that `geo_type` should become `column_type` and if it can be done it should remove some part of the code that alter table for geo type columns.

lmignon: You can take a look at https://github.com/acsone/alfodoo/blob/10.0-dev/cmis_field/fields/cmis_folder.py#L61

yvaucher: To note that PostGIS 2.0 allows to use `ALTER TABLE my_table ADD COLUMN 
geom geometry(Point, 4326);` which could fit in the `column_type`
Though the index should still be handmade.
https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L2512
I see no entry there to replace `gist_index` by `index` and to override creation of index.

And I fear that this monolithic method doesn't allow use to add the checks we already have in column update.

yvaucher:
To go further let's merge this base. no time yet to improve it but this is a working state.

PR #131, opened 2017-02-17
Update travis setup

yvaucher:


grindtildeath (changes requested):
I guess this could be closed or reworked ? :thinking:

Issue #132, opened 2017-02-20
[10.0] TypeError: Reprojection of column is not implementedWe can not change srid 900913 to 0


rygar08:
When installing 'Geo spatial support Demo' on odoo 10, i get this error....
TypeError: Reprojection of column is not implementedWe can not change srid 900913 to 0

Traceback (most recent call last):

.... traceback errors removed for brevity

  File "/home/parallels/cxn/odoo-10-cxn/odoo/modules/registry.py", line 315, in init_models
    model._auto_init()
  File "/home/parallels/cxn/oca/geospatial/base_geoengine/geo_model.py", line 55, in _auto_init
    fct(cr, f_name, self._table, self._name)
  File "/home/parallels/cxn/oca/geospatial/base_geoengine/fields.py", line 145, in update_geo_column
    self.srid, check_data[0]))
TypeError: Reprojection of column is not implementedWe can not change srid 900913 to 0

lesdekock:
Have found that the issue is with the Odoo core process creating the geom column instead of leaving it for the geo_model logic to create the field in the database. 

In the GeoModel _auto_init (see base_geoengin/geo_model.py) at line 41, 2 list are made, one for the base fields and one for the geo fields. 
On line 46 self._fields is equated to base_fields and then the super call is done to the _auto_init in the BaseModel (see odoo/models.py),

Here on line 2318 (in odoo/models.py) self = self.with_context(prefetch_fields=False) has the undesired effect of re-including the geo fields in self._fields and then the column is created by Odoo. 

Hope this helps in finding an amicable solution to the problem. For myself I just made a patch to drop any geo column created by Odoo with a srid = 0 and type = 'GEOMETRY' so that the geo_model can correctly recreate the column. 

Here is my patch in base_geoengin/geo_model.py @ line 47. 

```

def _auto_init(self):
    ....
    self._fields = base_fields
    res = super(GeoModel, self)._auto_init()
    ## Patch START ############################
    """Check if the column was created by Odoo core
       and DROP the column.
       if srid = 0 and type = 'GEOMETRY' then drop
    """
    for f_name in geo_fields.keys():
        query = ("""SELECT srid, type, coord_dimension
                 FROM geometry_columns
                 WHERE f_table_name = %s
                 AND f_geometry_column = %s""")
        cr.execute(query, (self._table, f_name))
        check_data = cr.fetchone()
        if check_data and check_data[0] == 0 and check_data[1] == 'GEOMETRY':
[...]

PR #134, opened 2017-04-05
[10.0][FIX] #132: geo columns truly manage by base geoengine

petrus-v:
mark fields type as manual let geogine the authority to create geom fields


petrus-v:
this is a purpose to fix #132 issue

yvaucher:
Thanks travis errors are unrelated and it does fix the issue #132

TypeError: Reprojection of column is not implementedWe can not change srid 900913 to 0

yvaucher (approved):

PR #150, opened 2017-08-10
[10.0] base_geoengine fix and improvements

benwillig:
* Implemented convert_to_cache and convert_to_record methods. (Store BaseGeometry hex code in cache and convert it back to BaseGeometry instance when reading from record)
* Added test in convert_to_read method to prevent errors when assigning a BaseGeometry instance to a Geo field.
* New method to create a GeoLine from two points
* Create Geometry column when column is computed but stored
* Fixed tests
* Corrected some pylint issues
* Allow to define which layers will be active on startup
* Allow to show a configurable popup when selecting an item.  Geonengine templates works like kanban box templates.
* Added a hook to allow to set a different style when selecting/hovering a layer
* Allow to change the circle radius with inheritance

### DIFF
--- a/base_geoengine/__manifest__.py
+++ b/base_geoengine/__manifest__.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     'name': 'Geospatial support for Odoo',
-    'version': '10.0.1.0.0',
+    'version': '10.0.2.0.0',
     'category': 'GeoBI',
     'author': "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
     'license': 'AGPL-3',
--- a/base_geoengine/__manifest__.py
+++ b/base_geoengine/__manifest__.py
@@ -23,8 +23,11 @@
      'security/ir.model.access.csv',
  ],
  'external_dependencies': {
-     'python': ['shapely',
-                'geojson'],
+     'python': [
+        'shapely',
+        'geojson',
+        'simplejson'
+     ],
  },
  'qweb': ["static/src/xml/geoengine.xml"],
  'installable': True,
--- a/base_geoengine/__manifest__.py
+++ b/base_geoengine/__manifest__.py
@@ -8,12 +8,11 @@
     'category': 'GeoBI',
     'author': "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
     'license': 'AGPL-3',
-    'website': 'http://openerp.camptocamp.com',
+    'website': 'https://github.com/OCA/geospatial',
     'depends': [
         'base',
         'web'
     ],
-    'init_xml': [],
     'data': [
         'security/data.xml',
         'views/base_geoengine_view.xml',
@@ -30,7 +29,9 @@
             'simplejson'
         ],
     },
-    'qweb': ["static/src/xml/geoengine.xml"],
+    'qweb': [
+        'static/src/xml/geoengine.xml',
+    ],
     'installable': True,
     'pre_init_hook': 'init_postgis',
 }
--- a/base_geoengine/__manifest__.py
+++ b/base_geoengine/__manifest__.py
@@ -2,34 +2,35 @@
 # Copyright 2011-2015 Nicolas Bessi (Camptocamp SA)
 # Copyright 2016 Yannick Vaucher (Camptocamp SA)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-{'name': 'Geospatial support for Odoo',
- 'version': '10.0.1.0.0',
- 'category': 'GeoBI',
- 'author': "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
- 'license': 'AGPL-3',
- 'website': 'http://openerp.camptocamp.com',
- 'depends': [
-     'base',
-     'web'
- ],
- 'init_xml': [],
- 'data': [
-     'security/data.xml',
-     'views/base_geoengine_view.xml',
-     'geo_ir/ir_model_view.xml',
-     'geo_view/ir_view_view.xml',
-     'geo_view/geo_raster_layer_view.xml',
-     'geo_view/geo_vector_layer_view.xml',
-     'security/ir.model.access.csv',
- ],
- 'external_dependencies': {
-     'python': [
-        'shapely',
-        'geojson',
-        'simplejson'
-     ],
- },
- 'qweb': ["static/src/xml/geoengine.xml"],
- 'installable': True,
- 'pre_init_hook': 'init_postgis',
- }
+{
+    'name': 'Geospatial support for Odoo',
+    'version': '10.0.1.0.0',
+    'category': 'GeoBI',
+    'author': "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
+    'website': 'http://openerp.camptocamp.com',
+    'depends': [
+        'base',
+        'web'
+    ],
+    'init_xml': [],
+    'data': [
+        'security/data.xml',
+        'views/base_geoengine_view.xml',
+        'geo_ir/ir_model_view.xml',
+        'geo_view/ir_view_view.xml',
+        'geo_view/geo_raster_layer_view.xml',
+        'geo_view/geo_vector_layer_view.xml',
+        'security/ir.model.access.csv',
+    ],
+    'external_dependencies': {
+        'python': [
+            'shapely',
+            'geojson',
+            'simplejson'
+        ],
+    },
+    'qweb': ["static/src/xml/geoengine.xml"],
+    'installable': True,
+    'pre_init_hook': 'init_postgis',
+}
--- a/base_geoengine/fields.py
+++ b/base_geoengine/fields.py
@@ -184,6 +184,30 @@ class GeoLine(GeoField):
     type = 'geo_line'
     geo_type = 'LINESTRING'
 
+    @classmethod
+    def from_points(cls, cr, point1, point2):
+        """
+        Converts given points in parameter to a line.
+        :param cr: DB cursor
+        :param point1: Point (BaseGeometry)
+        :param point2: Point (BaseGeometry)
+        :return: LINESTRING HEX
+        """
+        sql = """
+        SELECT
+            ST_MakeLine(
+                ST_GeomFromText(%(wkt1)s, 4326),
+                ST_GeomFromText(%(wkt2)s, 4326)
+            )
+        """
+        cr.execute(sql, {
+            'wkt1': point1.wkt,
+            'wkt2': point2.wkt,
+            'srid': cls._slots['srid'],
+        })
+        res = cr.fetchone()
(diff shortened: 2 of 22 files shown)


base_geoengine/fields.py
```diff
@@ -56,9 +57,22 @@ def convert_to_column(self, value, record):
...
 
+    def convert_to_cache(self, value, record, validate=True):
+        val = value
+        if isinstance(value, BaseGeometry):
+            val = value.wkb_hex
+        return val
+
+    def convert_to_record(self, value, record):
+        if value:
+            return self.load_geo(value)
```
lmignon: @benwillig ?`else`?

base_geoengine/fields.py
```diff
@@ -170,6 +184,30 @@ class GeoLine(GeoField):
...
+        Converts given points in parameter to a line.
+        :param cr: DB cursor
+        :param point1: Point (BaseGeometry)
+        :param point2: Point (BaseGeometry)
+        :return: LINESTRING HEX
+        """
+        sql = """
+        SELECT
+            ST_MakeLine(
+                ST_GeomFromText(%(wkt1)s, 4326),
```
lmignon: @benwillig Why do you hard code the `srid` value to use in the call to `ST_GeomFromText`?

lmignon (changes requested):

base_geoengine/fields.py
benwillig: I changed my code. I also discovered a new case which happen during onchange methods. When a onchange computed a GeoPoint, given value in convert_to_record method corresponds to the value returned by the convert_to_read method.

base_geoengine/fields.py
benwillig: Srid was hardcoded in ```GeoPoint.from_latlon``` method, so I thought there was a reason for that. Should I change the hardcoded ```srid``` in this method too?

base_geoengine/__manifest__.py
```diff
@@ -2,31 +2,35 @@
...
+    'version': '10.0.1.0.0',
+    'category': 'GeoBI',
+    'author': "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
+    'website': 'http://openerp.camptocamp.com',
+    'depends': [
+        'base',
+        'web'
+    ],
+    'init_xml': [],
```
dreispt: Deprecated, remove.

base_geoengine/__manifest__.py
```diff
@@ -2,31 +2,35 @@
...
+        'security/ir.model.access.csv',
+    ],
+    'external_dependencies': {
+        'python': [
+            'shapely',
+            'geojson',
+            'simplejson'
+        ],
+    },
+    'qweb': ["static/src/xml/geoengine.xml"],
```
dreispt: Insn't this deprecated in favour of a extending the assets with QWeb XML?

base_geoengine/__manifest__.py
```diff
@@ -2,31 +2,35 @@
...
- 'installable': True,
- 'pre_init_hook': 'init_postgis',
- }
+{
+    'name': 'Geospatial support for Odoo',
+    'version': '10.0.1.0.0',
+    'category': 'GeoBI',
+    'author': "Camptocamp,ACSONE SA/NV,Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
+    'website': 'http://openerp.camptocamp.com',
```
dreispt: This should really be the Github repo address, or a link to the docs.

base_geoengine/geo_model.py
```diff
@@ -42,12 +42,13 @@ def _auto_init(self):
         res = super(GeoModel, self)._auto_init()
         column_data = self._select_column_data()
         for f_name, geo_field in geo_fields.iteritems():
-            # XXX check if not computed field
-            if not geo_field.compute:
-                fct = geo_field.create_geo_column
-                if f_name in column_data:
-                    fct = geo_field.update_geo_column
-                fct(cr, f_name, self._table, self._name)
+            # XXX check if not computed stored field
```
dreispt: I suggest using `TODO` in place on `XXX`

base_geoengine/security/data.xml
```diff
@@ -1,15 +1,14 @@
...
-        </record>
+    <record id="group_geoengine_user" model="res.groups">
+        <field name="name">Geoengine user</field>
+    </record>
 
-    </data>
-</openerp>
\ No newline at end of file
+    <record id="group_geoengine_admin" model="res.groups">
+        <field name="name">Geoengine admin</field>
```
dreispt: s/admin/Admin

base_geoengine/__manifest__.py
benwillig: Not for qweb I think, unless I missed something. You extend the asset template only to add javascript and css file.

base_geoengine/geo_model.py
benwillig: I deleted this statement because I already fixed it.

base_geoengine/security/data.xml
benwillig: I don't understand what you mean.

JonathanNEMRY: @benwillig  I think the idea is to capitalize de "a" of "admin" (s/old_pattern/new_pattern) :smile:

benwillig: Thanks!

base_geoengine/__manifest__.py
lasley: * Not deprecated for static QWeb. Even if extending a static QWeb, you would need to include in here except in some special cases where the JS loads it

dreispt: ok

base_geoengine/fields.py
```diff
@@ -170,6 +191,30 @@ class GeoLine(GeoField):
...
+        SELECT
+            ST_MakeLine(
+                ST_GeomFromText(%(wkt1)s, %(srid)s),
+                ST_GeomFromText(%(wkt2)s, %(srid)s)
+            )
+        """
+        cr.execute(sql, {
+            'wkt1': point1.wkt,
+            'wkt2': point2.wkt,
+            'srid': cls._slots['srid'],
```
lmignon: @benwillig `srid` can differ by instance. IMO it would be better to add `srid=None` as parameter and if not specified, takes the default one on the class.

lmignon (changes requested):

lmignon:
@benwillig Thank you for all these improvements. I need to  functionally test all this to fully understand the scope of all these changes but these look promising when reading the code. (You can maybe update the documentation with an example of custom popup..) To ease the review and if you plan to add new functionalities, do it in new PR's based on this one.

benwillig:
@yvaucher ping

woodbrettm:
Hi there, any updates on this PR?

base_geoengine/static/src/js/views/geoengine_template_widgets.js
```diff
@@ -0,0 +1,97 @@
+/*---------------------------------------------------------
+ * Odoo base_geoengine
+ * Contributor N. Bessi Copyright Camptocamp SA
```
lmignon: @benwillig If all this code is new, you could remove N. Bessi as contributor and the Copyright for Camptocamp SA

lmignon (approved):
LGTM (code review)

base_geoengine/__manifest__.py
```diff
@@ -2,31 +2,36 @@
...
-     'python': ['shapely',
-                'geojson'],
- },
- 'qweb': ["static/src/xml/geoengine.xml"],
- 'installable': True,
- 'pre_init_hook': 'init_postgis',
- }
+{
+    'name': 'Geospatial support for Odoo',
+    'version': '10.0.1.0.0',
```
yvaucher: This deserve a version bump

base_geoengine/static/src/js/views/geoengine_template_widgets.js
yvaucher: Indeed 
Edit: was replying to @lmignon saying you have to credit only yourself here.

yvaucher (approved):
Nice work @benwillig 

Thanks for the cleaning and the improvements

"Geonengine templates works like kanban box templates." :heart:

yvaucher:
@benwillig can you just bump the version of the module and change credits so we can merge ?

base_geoengine/__manifest__.py
lasley: IMO a major version, so 10.0.2.0.0.

benwillig:
Thanks all for your reviews! I updated the version number of the module.

benwillig:
Should I fix the pylint error which cames from ```base_geoengine_demo```? Then I think this PR can be merged if there is no other questions/errors.

yvaucher:
@benwillig 
Here Is see that `base_geoengine` tests don't even start
https://travis-ci.org/OCA/geospatial/jobs/314630058

Can you fix the following error ?

```
MissingError: ('Error, can not automatically initialize spatial postgis support. Database user may have to be superuser and postgres/postgis extentions with their devel header have to be installed. If you do not want Odoo to connect with a super user you can manually prepare your database. To do thisopen a client to your database using a super user and run: \nCREATE EXTENSION postgis;\nCREATE EXTENSION postgis_topology;\n', None)
```

This error seems new in this PR.

For `base_geoengine_demo` no need to fix them here.

benwillig:
Unfortunately, I'm not sure if I can fix it. To create an extension in postgres, the connected user must be a superuser. The only way to fix it is to change the runbot instance creation. I don't know how it was working in previous versions.

yvaucher:
@benwillig it might be a step missing in the .travis.yml

Not sure either why those SQL requests aren't executed with a user which is not superuser. Could be due to version of travis/postgres/MQT

MQT doesn't seems to use a specific DB user:
https://github.com/OCA/maintainer-quality-tools/blob/master/travis/test_server.py#L233

We might be able to fix it by altering `travis` pg user to be superuser.

Could be worth just to do a commit --amend on last commit to trigger a new travis build.

benwillig:
@yvaucher I forced a rebuild but it didn't help. Indeed, if the travis user is used for the DB connection, it should be a postgres superuser.

benwillig:
Any news @yvaucher ?

yvaucher:
@benwillig reading https://docs.travis-ci.com/user/database-setup/#PostgreSQL
 
in .travis.yml try to add:

```
before_script:
  - psql -U postgres -c "create extension postgis"
```

benwillig:
Thanks for your feedback @yvaucher. It works now.

yvaucher:
I would have expected this to work.

Could you try to follow the recommendation here ?

Seems we have the same issue.

https://gis.stackexchange.com/questions/252609/could-not-access-file-libdir-postgis-2-3-on-travis-installation

benwillig:
I think you checked the first job. After it I found the link you gave and I added the postgis package to install in yml package category. The following job suceeded https://travis-ci.org/OCA/geospatial/jobs/367672576

yvaucher:
@benwillig you are right I didn't update this page, it was out of sync.

Thank you for having fixed this, I will then merge it!

PR #155, opened 2017-10-02
[10.0][MIG] geoengine_base_geolocalize migration

benwillig:
Also moved the openstreetmap api call in specific module.

@lmignon Your review is welcome!


ThomasBinsfeld (approved):

base_geolocalize_openstreetmap/models/geolocalize_openstreetmap.py
```diff
@@ -0,0 +1,45 @@
...
+    _name = 'geoengine.geolocalize.openstreetmap'
+
+    @api.model
+    def _geocode_address(
+            self, street, zip_code, city, state, country_name,
+            country_codes):
+        """Get the latitude and longitude by requesting "mapquestapi"
+        see http://open.mapquestapi.com/geocoding/
+        """
+        url = 'http://nominatim.openstreetmap.org/search'
```
JonathanNEMRY: as it is an openstreetmap class then could it make sens to externalize this one into a class attribute `_url` ?? (in order to reuser request into other call?)

JonathanNEMRY (approved):
Just a litle question otherwise seems ok

base_geolocalize_openstreetmap/models/geolocalize_openstreetmap.py
benwillig: Yes, you are right. I also thought about using a config parameter but I think it doesn't make sense as this URL will maybe never change.

lmignon (approved):
Works like a charm in production. Thank you @lmarion-source

PR #163, opened 2018-05-17
[9.0][FIX] Update map.lausanne.ch URI

p-tombez:
Change http://map.lausanne.ch/main/tiles/ to https://map.lausanne.ch/tiles/


Tonow-c2c (commented):
Ok for this may be take time to see what wrong for CI

guewen (approved):

PR #164, opened 2018-05-17
[10.0][FIX] Update map.lausanne.ch URI

p-tombez:
Change http://map.lausanne.ch/main/tiles/ to https://map.lausanne.ch/tiles/


pedrobaeza:
Why is this red?

Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

PR #166, opened 2018-06-28
[FIX] typo in base_geoengine javascript.

ecino:
There is a typo in the javascript of base_geoengine module which is causing Odoo to display a blank page in the backend. This fixes it.


yvaucher (approved):

yvaucher:
Thanks!

PR #167, opened 2018-07-17
[ADD] 8.0 base_geoengine: backport base_zoom from 9.0

mpanarin:
Backports of several improvements on base_geoengine from version 9.0:
1. Default zoom (can be specififed in the view) to open map on a specific zoom instead of zooming everytime by hand
2. Open related record - allows opening of the record connected to the geomap shape
3. Restricted extent and projection - allows change of projection and bound the map to an extent

### DIFF
--- a/base_geoengine/__init__.py
+++ b/base_geoengine/__init__.py
@@ -19,13 +19,13 @@
 #
 ##############################################################################
 """The GeoEngine module"""
+from . import geo_point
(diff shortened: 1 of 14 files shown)


yvaucher (approved):

yvaucher:
Merging as backports, and travis is the "normal" status.

PR #168, opened 2018-07-17
[8.0] [ADD] base_geoengine: backport labels on shapes

mpanarin:
backport of labels on shapes feature from 10.0


base_geoengine/static/src/js/geoengine_view.js
```diff
@@ -252,11 +252,27 @@ openerp.base_geoengine = function(openerp) {
             var features = [];
             var geostat = null;
             var geojson = new OpenLayers.Format.GeoJSON();
+            var l = data.length;
+            var f = data[0];
+            var v;
+            if (cfg.display_polygon_labels === true) {
+                v = cfg.attribute_field_id[1];
```
yvaucher: We need 2 different relations to a field, one for colors and a second one for labels.
Colors are defined depending on a value, while the label could be something completely different like a reference.

PR #170, opened 2018-08-21
[IMP] #4546 Now the partner is opened in a popup

gfcapalbo:
In the 9.0 main branch when you click on a partner it will show you the formview, but when you go back to the map your zoom and position is lost.
We had a client that thought this behavior was inconvenient, he needed to see in sequence many customers of the same area and every time he opened one , he would have to zoom in again.
Here, customer is loaded in popup form , so that the map remains intact.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,11 @@ sudo: false
 cache: pip
 
 addons:
-  postgresql: 9.6
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
       - postgis postgresql-9.6-postgis-2.3 postgresql-9.6-postgis-2.3-scripts # because travis doesn't know which one to install
-before_script:
-  - psql -U postgres -c "create extension postgis"
 
 language: python
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 cache: pip
 
 addons:
-  postgresql: 10
+  postgresql: 9.6
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,14 @@ sudo: false
 cache: pip
 
 addons:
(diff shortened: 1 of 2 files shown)


dek-odoo:
Good one..

gfcapalbo:
@dek-odoo thanks. 
this mod is very usable for users who need to make many calls / see many clients in a zoomed area. If we loose our map location on every click, it becomes difficult to work with map view.
The CI error is due to the fact that DB needs postgis.

yvaucher:
I understand the need for this proposal, nevertheless I see 2 issues:

1. the geoengine view is designed to act like the list view. Thus, openning a form should be on full page.
2. This PR seems only valid for partner object. You can set geometry on any kind of other models.

Instead of changing the default behavior it would be nice to have an option to set the "target: 'new'" behavior. IIRC there is also an option on tree view to define which view will be openned on click.

yvaucher (changes requested):
Tagging Request changes (see my previous comment)

gfcapalbo:
@yvaucher  i agree that  optional target:new would be a great option. I could make that a property?
the need for this was to increase usability, when the user would check out a partner and return, zoom would be lost.

Last week i had the need to make this work in a new window, not a popup. but investigating the web client I see that only current (same page) and new (popup) are supported.
Have you encountered such problem? would i have to submit an MR on web module to support that?

yvaucher:
> Last week i had the need to make this work in a new window

IMO forcing to open a new window sounds like a pop up hell. It should be the choice of the user to know where he wants to open it like with standard links.

Currently odoo doesn't even allow you to do a right click on an element of a view list to open it in a new tab or window as it is a js even and not an url link. We have the same issue with geoengine as we use the same js method.

What I haven't tested is to create a field using the widget url and forge the url to link you to the object I want to open. But that would be only a work-around. 

Would be great if odoo was listening on list views to ctrl+click or right click context menu option "Open in a new tab" to open in a new tab. We could do the same on the info box of geoengine.

PR #179, opened 2018-11-28
[FIX] Remove mode/format='hex'

mileo:
Signed-off-by: Luis Felipe Mileo <mileo@kmee.com.br>

Without this fix i'm having this error with:

- Chromium Versão 70.0.3538.77 (Versão oficial) Arch Linux 64 bits
- Firerox 63.0.1 (64-bit)

image


yvaucher:
Thanks, seems we did the same fix in 11.0

PR #182, opened 2019-01-21
Fix js error

mvrodriguez:


yvaucher:
I merged #179 instead

Thanks for reporting it.

PR #185, opened 2019-02-25
[MIG] base geoengine to 12.0

yvaucher:
Only few changes are required to make it work.

I'll push the portage of base_geoengine_demo in another PR

PR #187, opened 2019-03-19
[12.0][ADD] Support of Swisstopo layers

asaunier:
This PR depends on https://github.com/OCA/geospatial/pull/185 and will be rebased when the latter is merged. The only additional commit is e15da3e

The layers creation method has been moved from ``GeoengineMixin`` to a distinct class that is explicitely instanciated by the widget or renderer, because it seems not possible to override a mixin using Odoo's ``include()``.

See also PR https://github.com/OCA/geospatial/pull/188 for Odoo 11.0


yvaucher:
@asaunier I just merged #185 , can you rebase?

(some lint added some conflicts to your changes)

asaunier:
@yvaucher Thanks, updated!

yvaucher:
Travis is red for lint.

For JS lint we use ESlint

You can add a similar file in `static/src/js`

https://github.com/OCA/geospatial/blob/12.0/base_geoengine/static/src/js/.eslintrc

swisstopo_geoengine/geo_view/geo_raster_layer.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class GeoRasterLayer(models.Model):
+    _inherit = 'geoengine.raster.layer'
+
+    raster_type = fields.Selection(selection_add=[('swisstopo', 'Swisstopo')])
+    layername = fields.Char('Layer Machine Name')
+    #time_dimension = fields.Char('Time Dimension')
```
yvaucher: To remove ?

yvaucher: To remove ?

yvaucher (changes requested):
LGTM but Travis is red for lint and there is a remaining commented line.

For JS checks we use ESlint and the config used by the OCA is the following:

https://github.com/OCA/pylint-odoo/blob/master/pylint_odoo/examples/.jslintrc

You can add a similar file in `static/src/js` :

https://github.com/OCA/geospatial/blob/12.0/base_geoengine/static/src/js/.eslintrc

BTW nice change about refactoring BackgroundLayer in a separate class.

swisstopo_geoengine/geo_view/geo_raster_layer.py
asaunier: Actually I have forgotten to add this field, it is supposed to be used in order to set the time dimension in the swisstopo URLs. It defaults to ``current``:
https://github.com/OCA/geospatial/blob/7f66a4e00fe1c30502d0f0efc23fdd04987ccec6/swisstopo_geoengine/static/src/js/swisstopo_geoengine.js#L15

yvaucher (changes requested):
After some test I get the following error on form view with the widget:

TypeError: this.rasterLayers[0] is undefined

http://localhost:8888/base_geoengine/static/src/js/widgets/geoengine_widgets.js:94

Errors happen in the following code in:

```
        _createLayers: function (field_infos) {
            this.vectorLayer = this._createVectorLayer();
            this.rasterLayers = this.bgLayers.create([
                field_infos.edit_raster,
            ]);
            this.rasterLayers[0].isBaseLayer = true;
        },
```

EDIT: In my test I had set a geometry with raster_type = 'swisstopo' but didn't loaded the module swisstopo_geonengine which should fallback to `default` background (empty).

base_geoengine/static/src/js/widgets/geoengine_widgets.js
```diff
@@ -86,7 +88,7 @@ odoo.define('base_geoengine.geoengine_widgets', function (require) {
 
         _createLayers: function (field_infos) {
             this.vectorLayer = this._createVectorLayer();
-            this.rasterLayers = this.createBackgroundLayers([
+            this.rasterLayers = this.bgLayers.create([
                 field_infos.edit_raster,
             ]);
             this.rasterLayers[0].isBaseLayer = true;
```
yvaucher: When we fallback on default (due to user error on raster layer definition) this.rasterLayers will be an empty list and fail.

asaunier:
@yvaucher Module renamed to ``geoengine_swisstopo`` as you suggested.

swisstopo_geoengine/geo_view/geo_raster_layer.py
```diff
@@ -10,4 +10,4 @@ class GeoRasterLayer(models.Model):
     raster_type = fields.Selection(selection_add=[('swisstopo', 'Swisstopo')])
     layername = fields.Char('Layer Machine Name')
     time = fields.Char('Time Dimension')
-    projection = fields.Char('Projection')
+    # projection = fields.Char('Projection')
```
yvaucher: Please remove it completely instead of commenting it. The change is quite simple and thus it has very few value to keep it.

swisstopo_geoengine/geo_view/geo_raster_layer_view.xml
```diff
@@ -14,10 +14,10 @@
                 <group string="Swisstopo WMTS options"
                        attrs="{'invisible': [('raster_type', '!=', 'swisstopo')]}"
                        colspan="4">
-                  <p colspan="2">See <a href="https://api3.geo.admin.ch/services/sdiservices.html#wmts" target="_blank">GeoAdmin API documentation</a> for registration info and available layers.</p>
+                  <p colspan="2">See <a href="https://api3.geo.admin.ch/services/sdiservices.html#wmts" target="_blank">GeoAdmin API documentation</a> for registration info and available layers. As for now only projection EPSG:21781 is supported.</p>
                   <field name="layername"/>
                   <field name="time"/>
-                  <field name="projection"/>
+                  <!--<field name="projection"/>-->
```
yvaucher: Same here

yvaucher (changes requested):

yvaucher:
Only those 2 last comment to change and we can merge.

asaunier:
@yvaucher Updated, bb48d52 addresses your suggestions.
I have also squashed some commits.

yvaucher (approved):

PR #188, opened 2019-03-19
[11.0][ADD] Support of Swisstopo layers

asaunier:
Similar to the changes suggested in 12.0, see https://github.com/OCA/geospatial/pull/187

The layers creation method has been moved from ``GeoengineMixin`` to a distinct class that is explicitely instanciated by the widget or renderer, because it seems not possible to override a mixin using Odoo's ``include()``.


yvaucher:
See fixes done in #187

yvaucher (changes requested):
TODO: backport fixes from #187

asaunier:
@yvaucher Updated. This PR is in sync with #187

yvaucher (approved):

PR #194, opened 2019-03-25
[10.0] Fix CI build adding postgis scripts

yvaucher:
Postgis scripts are included in travis build.
However, runbot doesn't pre-install them. Thus it needs to be set
explicitely.


yvaucher:
Merging, there are still some lint errors but unit tests pass and runbot too.

PR #196, opened 2019-04-09
[12.0][MIG] base_geoengine_demo

yvaucher:
Demo module migrated to version 12.0

Here are the 2 geoengine views for both demo objects (ZIP and Retail machines)

geoengine_demo_12-1
geoengine_demo_12-2

And here is an in-form widget

geoengine_demo_12-3


base_geoengine_demo/views/retail_machine.xml
```diff
@@ -110,37 +101,37 @@
...
-  <field name="geo_field_id" ref="base_geoengine_demo.field_geoengine_demo_automatic_retailing_machine_the_point"/>
+  <field name="geo_field_id" ref="base_geoengine_demo.field_geoengine_demo_automatic_retailing_machine__the_point"/>
   <field name="name">Retail Machine State colored</field>
   <field name="classification">unique</field>
   <field eval="6" name="sequence"/>
   <field name="view_id" ref="ir_ui_view_retailmachineview0"/>
   <field name="geo_repr">colored</field>
   <field eval="1" name="nb_class"/>
-  <field name="attribute_field_id" ref="base_geoengine_demo.field_geoengine_demo_automatic_retailing_machine_state"/>
+  <field name="attribute_field_id" ref="base_geoengine_demo.field_geoengine_demo_automatic_retailing_machine__state"/>
```
Tonow-c2c, line 111: I didn't understand why you add a new `_` before `_state`

yvaucher: This is new in Odoo 12.0, xmlids of fields, now we have 2 underscores between the name of the model and the name of the field.

Tonow-c2c: ok thanks

Tonow-c2c (approved):
LGTM Just code review

Sixologue:
why the value of the attributes on the map_info_box does not support the modified value ?

gitlcp:
base_geoengine_demo   Installation failure
TIM图片20190713053438

yvaucher:
@gitlcp in your case it is like the field https://github.com/OCA/geospatial/pull/196/files#diff-34e89c563984334a4659419c0157ac09R15 wasn't created or is not found.

I have a doubt about your instance, your path is odoo11/odoo12.

If you are on a Odoo 11.0 instance it will fail because auto generated XML ids (ir.model.data) for fields have changed with an extra underscore that separate the model name from the field name.

```
base_geoengine_demo.field_dummy_zip_the_geom (11.0)
base_geoengine_demo.field_dummy_zip__the_geom (12.0)
```

EDIT:

just saw https://github.com/OCA/geospatial/issues/206

yvaucher:
@Sixologue you might want to open an issue about this? I guess it is more related to base_geoengine than to the demo module. Or am I missing something?

EDIT: https://github.com/OCA/geospatial/issues/198

PR #218, opened 2019-09-30
Fix state of field for geo field to 'base'

yvaucher:
It was previously state `manual`
Usage of `manual` is for custom fields created from the interface such
as fields prefixed by `x_`.
In commit odoo/odoo@547118d manual field don't have an xmlid anymore.

Thus it had the side effect to fail when loading definition of vector layers
from an XML file as an XMLID is required to reference the field.
The XMLID would not be found.

### DIFF
--- a/base_geoengine/fields.py
+++ b/base_geoengine/fields.py
@@ -39,7 +39,6 @@ class GeoField(fields.Field):
         'dim': 2,
         'srid': 3857,
         'gist_index': True,
-        'manual': True,
     }
 
     def convert_to_column(self, value, record, values=None):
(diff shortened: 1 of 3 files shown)


tschanzt (approved):

tschanzt:
@yvaucher looks good to me, but travis is red

max3903 (approved):

yvaucher:
@max3903 I updated it and travis is green now.

yvaucher:
FYI @petrus-v 

2f46bad77b845819139115e0e89938f2c5e0f74f

PR #222, opened 2019-10-15
[ADD][12.0] google_maps

murtuzasaleh:
WIP

PR #223, opened 2019-10-15
[ADD][12.0] web_view_google_map

murtuzasaleh:
Depends on:

- [x] #228 base_google_map


pedrobaeza:
I don't think the module should be name with that general name, and it also seems to contain several things that can be split.

cc @yvaucher

murtuzasaleh:
#222

wolfhall:
@pedro, we thought about the name for this and Google Maps seemed to be most intuitive for what someone would expect from a tool that offers Google maps and Google mapping API services. 
What would you name this?

pedrobaeza:
Well, I don't know enough your module, but several patterns:

- It should start with the main area you are modifying (as this is not a new area like field-service). In this context, it should be `web_...`.
- It adds a new view (plus some widgets that I consider should be split in other modules, but maybe it's not possible/desirable), so we should continue with the name with `_view_`.
- Plurals must be avoided according OCA guidelines (https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules), so end it with `_google_map`.

Any way, I have also the doubt of why not using existing map view in `base_geoengine` (or extend it). That's why I pong @yvaucher, which is the most suitable for telling.

dreispt:
First of all, congratulations, @murtuzasaleh this is top work!
I can say it's far ahead my personal Odoo development skills.

Having said that, I agree with some of Pedro's points.

1. This mainly adds a new webclient view mode. So the intuitive name for it would be `web_view_google_map` or `web_view_googlemap` . I still think that this repo is more suited to host this module, than OCA/web.

2. It does more than that, and also adds two widgets and an autocomplete function. If these features are so tightly related that this becomes hard, I also suggest splitting those features in three or four modules, and have a master one, `web_googlemaps`, that adds the common configuration settings, and depends on the others. The goal here is that smaller modules are easier to maintain. Whoever this is not a showstopper.

3. Regarding extending `base_geoengine`, I believe that the difference here is that this map implementation does not require PostGIS.

From a code review PoV, I code looks good.
I'll be testing this and will post a review.

pedrobaeza:
> Regarding extending base_geoengine, I believe that the difference here is that this map implementation does not require PostGIS.

Then maybe a refactoring over `base_geoengine` that allows to reuse the same base for both kind of views: `web_view_map` or similar. You can start creating it setting a map provider agnostic layer, and adapt it in a second module for Google Maps, and then Geoengine can later be adapted to this base (maybe in 13.0).

wolfhall:
I like the idea of re-architecturing, but right now we have no budget and a customer who is waiting. 
Maybe we can have this as a roadmap.

There is a hierarchy for several sub-modules under google map. In order to be consistent with the Geospatial naming (base_genengine) we could go with base_google_map as the top-level module and then have related component modules underneath.
**base_google_map**
- web_view_google_map
- web_widget_google_marker_icon_picker
- web_widget_google_address_form  

Will that work better?

pedrobaeza:
Yeah, that sounds good, and I understand the budget issue. One question, what `base_google_map` contains?

dreispt:
That should be the configuration settings plus the dependencies on the "component modules". It's the same idea I proposed in my bullet 2.

pedrobaeza:
OK, understood. I suppose credentials, token and so on. Good for me. Please annotate in ROADMAP about the extraction of `web_view_map`.

wolfhall:
Will do!  Great comments.

google_maps/__manifest__.py
```diff
@@ -0,0 +1,39 @@
+# Copyright (C) 2019, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Google Maps',
```
dreispt: Odoo 12 includes a module with exactly the same name, bit for a different feature (add maps to the website).
I suggest expanding the title.
For example "Google Maps Features"

google_maps/README.md
```diff
@@ -0,0 +1,289 @@
+# Web Google Maps
```
brian10048: README will need to be broken up into separate files in the readme directory

google_maps/README.md
```diff
@@ -0,0 +1,289 @@
...
+        ...
+        <field name="view_type">form</field>
+        <field name="view_mode">tree,form,map</field>
+        ...
+    </record>
+```
+
+The view looks familiar?    
+Yes, you're right.    
+The marker infowindow will use `kanban-box` kanban card style.
```
brian10048:
```suggestion
The marker info window will use `kanban-box` kanban card style.    
```

google_maps/views/res_partner.xml
```diff
@@ -0,0 +1,104 @@
...
+                                </ul>
+                                <div class="oe_kanban_partner_links"/>
+                            </div>
+                        </div>
+                    </t>
+                </templates>
+            </map>
+        </field>
+    </record>
+    <record id="base.action_partner_form" model="ir.actions.act_window">
```
brian10048: I expected the Map view to be an option when going to Contacts application.  

I had to go to Sales app > Orders > Customers for the Map view to be available.

brian10048 (changes requested):
Awesome work!

google_maps/views/res_partner.xml
dreispt: @brian10048 I found the same issue and investigated it . It turns out that "Contacts" rolls out its own Window Action, while Sale and CRM reuse the Partner's Window Action defines in the Base module, extended in this module.
To make it work for Contacts we would have to add it as a dependency, which I think is not a good thing.

google_maps/views/res_partner.xml
```diff
@@ -0,0 +1,104 @@
...
+        <field name="view_mode">kanban,tree,form,map</field>
```
dreispt: Activity view is missing
```suggestion
        <field name="view_mode">kanban,tree,form,activity,map</field>
```

dreispt:
I'm excited with this contribution.
I ran tests locally and it's working good.

Having the module split would be great.
I suggest:

* `base_google_map` (make it an app, hold the configuration settings). Ideally it would depend/ install the Map view. and would add options to install the widgets.
  * web_view_google_map`
  * `web_widget_google_address_autocomplete`
  * `web_widget_google_partner_autocomplete`

Thanks!

murtuzasaleh:
> First of all, congratulations, @murtuzasaleh this is top work!
> I can say it's far ahead my personal Odoo development skills.
> 
> Having said that, I agree with some of Pedro's points.
> 
> 1. This mainly adds a new webclient view mode. So the intuitive name for it would be `web_view_google_map` or `web_view_googlemap` . I still think that this repo is more suited to host this module, than OCA/web.
> 2. It does more than that, and also adds two widgets and an autocomplete function. If these features are so tightly related that this becomes hard, I also suggest splitting those features in three or four modules, and have a master one, `web_googlemaps`, that adds the common configuration settings, and depends on the others. The goal here is that smaller modules are easier to maintain. Whoever this is not a showstopper.
> 3. Regarding extending `base_geoengine`, I believe that the difference here is that this map implementation does not require PostGIS.
> 
> From a code review PoV, I code looks good.
> I'll be testing this and will post a review.

@dreispt  Thank you, Module development credit goes to @gityopie

murtuzasaleh:
> I'm excited with this contribution.
> I ran tests locally and it's working good.
> 
> Having the module split would be great.
> I suggest:
> 
> * `base_google_map` (make it an app, hold the configuration settings). Ideally it would depend/ install the Map view. and would add options to install the widgets.
>   
>   * web_view_google_map`
>   * `web_widget_google_address_autocomplete`
>   * `web_widget_google_partner_autocomplete`
> 
> Thanks!

Ok, I will work on it and create PR based on that.

murtuzasaleh:
> I like the idea of re-architecturing, but right now we have no budget and a customer who is waiting.
> Maybe we can have this as a roadmap.
> 
> There is a hierarchy for several sub-modules under google map. In order to be consistent with the Geospatial naming (base_genengine) we could go with base_google_map as the top-level module and then have related component modules underneath.
> **base_google_map**
> 
> * web_view_google_map
> * web_widget_google_marker_icon_picker
> * web_widget_google_address_form
> 
> Will that work better?

@dreispt @wolfhall I will rename google_map module to web_view_google_map.

dreispt:
Does that mean that you don't plan to split the features?
If we stay with a single module, I'm not sure if we shouldn't rather use `base_google_maps`, for consistency with `base_geoengine`.

Thoughts?

murtuzasaleh:
> Does that mean that you don't plan to split the features?
> If we stay with a single module, I'm not sure if we shouldn't rather use `base_google_maps`, for consistency with `base_geoengine`.
> 
> Thoughts?

@dreispt So in the `base_google_map` module will create separately so all the map (.js code and library) and configuration functionality will go under the `base_google_map` module and only display **partner demo map view** functionality go under the `web_view_google_map` or the `web_view_google_map_contacts` module?

Is it the same as `base_geoengine` and `base_geoengine_demo` modules?

Please correct me if I am wrong?

dreispt:
That was not what I meant:

- `base_google_map`:  provides the configuration settings, and dependency-installs the other components.
- `web_view_google_map`: provides the js+css+xml, etc. for the Map view
- `web_widget_google_address_autocomplete`: code for the address autocomplete widget
- ` web_widget_google_partner_autocomplete`: code for the partner data autocomplete widget

You could also add:
- `base_google_map_demo`: add the Google Map features to the Contacts app.

murtuzasaleh:
> * base_google_map

@dreispt I have discussed with @wolfhall and for now, I will create PR based on these points.

1) base_google_map (the parent module in the Geospatial Repo)
2) web_view_google_map (renamed google_map module)
3) web_widget_ google_marker_icon_picker
4) web_widget_google_address_autocomplete

dreispt:
OK, upon code inspection I came up with a different list of implemented widgets, but I could have got that wrong, so please go ahead.

The good practice for this is to have one PR per module proposed.
Thank you.

murtuzasaleh:
#228

murtuzasaleh:
@dreispt Can you please review this PR.

web_view_google_map/README.rst
```diff
@@ -0,0 +1,358 @@
+===============
```
dreispt, line 1: To avoid the RST lint errors, copy the template at https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

web_view_google_map/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2019, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Google Map View',
```
dreispt, line 5:
```suggestion
    'name': 'Google Map View',
    'summary': 'Add a Google Map view type to the Odoo web client',
```

dreispt (commented):
LGTM, please fix Travis

murtuzasaleh:
Depend on #228

dreispt:
I configured an API key on Runbot (http://3397005-223-7fc689.runbot1-2.odoo-community.org) and did functional tests, and it's working good! 

image

dreispt (approved):

brian10048 (approved):
I tested on the runbot as well

PR #236, opened 2019-11-22
[12.0][IMP] Improved code.

murtuzasaleh:


web_widget_google_marker_icon_picker/static/src/js/view/map/map_renderer.js
```diff
@@ -7,14 +7,8 @@ odoo.define('web_widget_google_marker_icon_picker.MapRenderer', function (requir
         _initLibraryProperties: function (params) {
             this._super.apply(this, arguments);
             if (this.mapLibrary === 'geometry') {
-                this.fieldMarkerColor = params.fieldMarkerColor;
                 this.iconUrl = '/web_widget_google_marker_icon_picker/static/src/img/markers/';
             }
         },
-        _createMarker: function (latLng, record, color) {
-            var color = (record.data[this.fieldMarkerColor] ? record.data[this.fieldMarkerColor] : color) || 'red';
-            this._super(latLng, record, color);
```
dreispt, line 16: Can we have an explanation about this change?
I see removing the ability to set a custom color, and instead limit markers to an available marker image file.
I'm curious because I see code removed but no code added.

murtuzasaleh: @dreispt No need for this code image marker will create based on https://github.com/OCA/geospatial/pull/235/files#diff-88d4fe3ec16c0db00836315ebd63b287R1047

dreispt: I'm wondering if this is relevant as a separate module now.
I'm giving my 👍 to this.

dreispt (approved):

PR #237, opened 2019-12-11
[12.0] base_geoengine: fix tab listener after pager

grindtildeath:
* When changing records using the pager the map wouldn't display due to not going through the event DOM_updated, thus we need an other way to add the tab listener in order to trigger the init of the map at the right time. Which is when the user clicks on a tab to open it. As only then we will know what is the dedicated space for the map.

* When displaying the map before editing it, a JS error was raised because the geometries layer were not loaded correctly on a tab switching action.

### DIFF
--- a/base_geoengine/static/src/js/views/form_renderer.js
+++ b/base_geoengine/static/src/js/views/form_renderer.js
@@ -1,0 +1,19 @@
+odoo.define('base_geoengine.form_renderer', function (require) {
+    "use strict";
+
+    var FormRenderer = require('web.FormRenderer');
+
+    FormRenderer.include({
+        // _updateView refreshes the view after a pager action.
+        // Add the TabListener to geo_type widget so that _renderMap is called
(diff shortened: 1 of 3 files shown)


yvaucher (approved):

yvaucher (approved):

base_geoengine/static/src/js/views/form_renderer.js
```diff
@@ -0,0 +1,19 @@
+odoo.define('base_geoengine.form_renderer', function (require) {
+    "use strict";
+
+    var FormRenderer = require('web.FormRenderer');
+
+    FormRenderer.include({
+        // _updateView refreshes the view after a pager action.
+        // Add the TabListener to geo_type widget so that _renderMap is called
```
yvaucher, line 8: We can probably keep `_renderMap` in this comment even if not completely accurate due to last change.

PR #239, opened 2020-01-13
[12.0] Fix travis - Ignore false positive lint check E8103 and fix for BlockingIOError

yvaucher:
Here the aim is to fix travis.

Currently pylint gives false positives
Once pylint-odoo accepts SQL string composition
we could rewrite the query.

Pylint-odoo issue tracked here:
https://github.com/OCA/pylint-odoo/issues/219

Those queries are low level and the parameters should be safe.

geo_search might need a double check.

Plus, probably related to numerous lint warnings on a module, we get a `BlockingIOError` (which doesn't appear on builds that targets only the modules that are more respectful to js lint guidelines)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,18 @@ env:
   - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
   - TESTS="1" ODOO_REPO="OCA/OCB"
 
+before_install:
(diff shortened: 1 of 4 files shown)


yvaucher:
Finally, got rid of BlockingIOError on @OCA-git-bot ...

PR #243, opened 2020-02-12
[13.0][IMP] Use default zoom when creating a map with no geometry

yvaucher:
Forward port of https://github.com/OCA/geospatial/pull/231


yvaucher:
Fixing lint in https://github.com/OCA/geospatial/pull/244

PR #246, opened 2020-02-28
[ADD] geoengine_bing: Aerial with labels on demand

tarteo:


yvaucher (approved):
According to this https://powerapps.microsoft.com/en-us/blog/image-control-static-maps-api/

It's indeed missing from the list.

PR #250, opened 2020-04-01
13.0 mig geoengine partner

Tonow-c2c:
Result :
image


geoengine_partner/views/partner.xml
```diff
@@ -40,7 +40,7 @@
 
     <record id="geoengine_vector_layer_res_partner" model="geoengine.vector.layer">
```
grindtildeath, line 41: Looks like this should be moved to a data file instead of partner views.

grindtildeath (approved):
Looks good :+1:

geoengine_partner/views/partner.xml
yvaucher: This is data to define the widget which is part of the view. IMO it's fine to have it here. It could otherwise be moved in a dedicated file for geo config of the view but still in the `views` folder.

yvaucher (approved):

simahawk:
@Tonow-c2c pls check pre-commit issues

p-tombez:
@simahawk The pre-commit hook is crashing, nothing to do with an actual linting issue as previous build was green

PR #251, opened 2020-04-01
[MIG] geoengine_base_geolocalize: Migration to 13.0

p-tombez:
Pre-requisites:
- [x] #250


geoengine_base_geolocalize/__manifest__.py
```diff
@@ -19,13 +19,13 @@
...
     "name": "Geospatial support for base_geolocalize",
-    "version": "11.0.0.1.0",
+    "version": "13.0.1.0.0",
     "category": "GeoBI",
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "website": "http://www.acsone.eu",
     "depends": ["base", "geoengine_partner", "base_geolocalize"],
-    "external_dependencies": {"python": ["requests"]},
+    "external_dependencies": {"python": ["requests>=2.23.0"]},
```
yvaucher: @p-tombez what is the reason behind limiting the version of `requests` to a newer version?

Odoo itself already requires `requests` but is expecting the `2.21.0`
https://github.com/odoo/odoo/blob/13.0/requirements.txt#L40

If there is no reason to have a version higher than `2.21.0` I would suggest to remove it from the dependencies.

p-tombez: There is no need to tie it up to a newer version (I didn't check the requirements from Odoo). Should I remove completely the `external_dependencies` part ?

geoengine_base_geolocalize/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+try:
+    import requests
+except ImportError:
+    logger = logging.getLogger(__name__)
+    logger.warning("requests is not available in the sys path")
+
+_logger = logging.getLogger(__name__)
+
+
+class ResPartner(geo_model.GeoModel):
```
yvaucher:
```suggestion
class ResPartner(models.Model):
```

geoengine_base_geolocalize/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        for partner in self:
+            if not partner.partner_latitude or not partner.partner_longitude:
+                partner.geo_point = False
+            else:
+                partner.geo_point = geo_fields.GeoPoint.from_latlon(
+                    partner.env.cr, partner.partner_latitude, partner.partner_longitude
+                )
+
+    geo_point = geo_fields.GeoPoint(readonly=True, store=True, compute="_get_geo_point")
```
yvaucher:
```suggestion
    geo_point = fields.GeoPoint(readonly=True, store=True, compute="_get_geo_point")
```

geoengine_base_geolocalize/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import logging
+
+from odoo import api, exceptions, fields
```
yvaucher:
```suggestion
from odoo import api, exceptions, fields, models
```

geoengine_base_geolocalize/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+from odoo.tools.translate import _
+
+from odoo.addons.base_geoengine import fields as geo_fields, geo_model
```
yvaucher:
```suggestion
```

yvaucher (changes requested):
Thanks for the migration.

After testing, we need to change the following.
Since v12 `GeoModel` should not be used, use `Model` instead.
Same for `geo_fields` which are now registered in standard `fields`

geoengine_base_geolocalize/__manifest__.py
yvaucher: Yes, I'm in favor of completely remove the `external_dependencies`.

geoengine_base_geolocalize/views/res_partner_view.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record model="ir.ui.view" id="geoengine_partner.geo_partner_view_form">
```
grindtildeath, line 3: `geoengine_partner` in the `id` looks odd :confused: 
IMO would be better to use it in `inherit_id` if the goal is to change the location of the field in the view...

geoengine_base_geolocalize/tests/test_geoengine_partner.py
```diff
@@ -0,0 +1,58 @@
...
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
```
grindtildeath, line 18: Shouldn't big headers have been removed in a previous migration? :thinking:

geoengine_base_geolocalize/models/res_partner.py
grindtildeath:
```suggestion
    geo_point = geo_fields.GeoPoint(readonly=True, store=True, compute="_compute_geo_point")
```

geoengine_base_geolocalize/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+                    "date_localization": fields.Date.today(),
+                }
+            )
+
+    def geo_localize(self):
+        self.geocode_address()
+        return True
+
+    @api.depends("partner_latitude", "partner_longitude")
+    def _get_geo_point(self):
```
grindtildeath:
```suggestion
    def _compute_geo_point(self):
```

geoengine_base_geolocalize/models/res_partner.py
```diff
@@ -0,0 +1,99 @@
...
+                "country": partner.country_id and partner.country_id.name or "",
+                "countryCodes": partner.country_id and partner.country_id.code or "",
+            }
+
+            request_result = requests.get(url, params=pay_load, headers=headers)
+            try:
+                request_result.raise_for_status()
+            except Exception as e:
+                _logger.exception("Geocoding error")
+                raise exceptions.Warning(_("Geocoding error. \n %s") % str(e))
```
grindtildeath:
```suggestion
                raise exceptions.UserError(_("Geocoding error. \n %s") % str(e))
```

grindtildeath (changes requested):
Small things

Tonow-c2c (approved):

yvaucher (approved):

yvaucher:
waiting on #250 (merge failed on it)

PR #255, opened 2020-05-08
[MIG] geoengine_swisstopo: Migration to 13.0

TDu:


Tonow-c2c (approved):

geoengine_swisstopo/i18n/geoengine_swisstopo.pot
```diff
@@ -0,0 +1,85 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* geoengine_swisstopo
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 12.0\n"
```
mmequignon, line 7: 13.0 ?

mmequignon (approved):
LG

geoengine_swisstopo/i18n/geoengine_swisstopo.pot
TDu: This is because I did not have to regenerate translations.
Not sure if it should be done.

yvaucher:
@TDu it seems you lost all the history of the module as it was rename 
from swisstopo_geoengine to geoengine_swisstopo

Could you include the history of swisstopo_geoengine?
Or at least rename the first commit? and mention that it is a squash of those commits:

https://github.com/OCA/geospatial/commits/0c1c91498ef4fba6a29d756bf2221c6752f163e0/swisstopo_geoengine


My main concern is mostly about loosing this commit:
https://github.com/OCA/geospatial/commit/db2084376070d7d36b185a14977407f61fb4ed0a#diff-c08e6332b32af5347c522b1605c7d735

So the best way would be to rewrite history to make sure the module was always named `geoengine_swisstopo`.

IIRC the renaming happened in the original PR.

TDu:
@yvaucher Ok so indeed I did not have the commit history from before the renaming of the module.
I also learnt that when you rename some file/directory (even with the proper git command) the commit history is not visible in Github anymore. Although it is still visible through the `git --follow` command.
To fix this I updated the patch so those commit are targetting the new filename and dropped the commit containing the rename of the module.

Issue #252, opened 2020-04-16
[12.0] Assertion failed. See https://openlayers.org/en/v3.20.1/doc/errors/#25 for details


MagnusKriel:
Specified a map view using this XML definition:
`  <record id="geo_data_point_map_view" model="ir.ui.view">
    <field name="name">Geo Data Point Map View</field>
    <field name="arch" type="xml">
      <geoengine>
          <field name="geo_point"/>
      </geoengine>
    </field>
    <field name="model">geo.data.point</field>
    <field name="default_extent">18344886.788442, -5943743.319455, 19954344.856015,  -4099470.700991</field>
  </record>`
The default_context I specify is a rectangle around New Zealand using EPSG:900913 projection.
When opening the view in edit mode to create a new record (no data point set as it is a new record) I get this error in the browser:
Assertion failed. See https://openlayers.org/en/v3.20.1/doc/errors/#25 for details.

The map then renders at some other location:
![image](https://user-images.githubusercontent.com/20549666/79442002-0cce8980-802c-11ea-9f7f-fd3693bd78a4.png)

If not specifying the default_context it works but default extent is over Europe.

MagnusKriel:
Found the cause of this issue:
in geoengine_widgets.js, line 121 (Odoo12 branch):
`_updateMapEmpty: function () {
            var map_view = this.map.getView();
            // Default extent
            if (map_view) {
                var extent = this.defaultExtent.split(', ');
                map_view.fit(extent, {maxZoom: this.defaultZoom || 5});
            }
        },`
The extent is a list of strings and when it is passed to the assert function for validation:
`ol.extent.isEmpty = function(extent) {
  return extent[2] < extent[0] || extent[3] < extent[1];
};`
It does a string compare instead of intended float compare.
This is why this extent fails:
`["18344886.788442", "-5943743.319455", "19954344.856015", "-4099470.700991"]`

yvaucher:
@MagnusKriel nice catch!

Shouldn't be too hard to fix.

For the default extent, would make sense to place it at 0,0 and move the europe default_extent in the geoengine demo module.

PR #257, opened 2020-05-19
[12.0] Fix default_extent in views

TDu:
Convert the extent information from string to number when opening an
empty view.
Although the default_extent is set with a default value on creation, the
field can be set to null so make sure a default value is used.

This is to attend https://github.com/OCA/geospatial/issues/252

### DIFF
--- a/base_geoengine/static/src/js/widgets/geoengine_widgets.js
+++ b/base_geoengine/static/src/js/widgets/geoengine_widgets.js
@@ -207,6 +207,9 @@ odoo.define('base_geoengine.geoengine_widgets', function (require) {
             // Default extent
             if (map_view) {
                 var extent = this.defaultExtent.split(', ');
+                Object.keys(extent).forEach(key => {
+                    extent[key] = Number(extent[key])
+                });
                 map_view.fit(extent, {maxZoom: this.defaultZoom || 5});
             }
         },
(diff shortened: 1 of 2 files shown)


sebalix (approved):

i-vyshnevska (approved):

base_geoengine/static/src/js/widgets/geoengine_widgets.js
```diff
@@ -207,6 +207,9 @@ odoo.define('base_geoengine.geoengine_widgets', function (require) {
             // Default extent
             if (map_view) {
                 var extent = this.defaultExtent.split(', ');
```
yvaucher: could we take the opportunity to be more resilient on this split?
```suggestion
                var extent = this.defaultExtent.replace(/\s/g, '').split(',');
```

base_geoengine/static/src/js/widgets/geoengine_widgets.js
```diff
@@ -207,6 +207,9 @@ odoo.define('base_geoengine.geoengine_widgets', function (require) {
...
+                Object.keys(extent).forEach(key => {
+                    extent[key] = Number(extent[key])
+                });
```
yvaucher: Using `map` might be simpler here.

```suggestion
                extent = extent.map(coord => Number(coord));
```

yvaucher (changes requested):
Thanks.

Here is a couple of improvement suggestions.

TDu:
@yvaucher Thanks for the suggestion, works well

yvaucher (approved):

PR #262, opened 2020-06-19
[10.0][FIX] demo data: unknown field 'symbol_binary' / fix travis

lmignon:


lmignon:
fast tracking to fix travis

PR #264, opened 2020-06-30
[ADD] base_geo_route_optimization

brian10048:
This module provides an extendable framework for solving the Traveling Salesman Problem and/or the Vehicle Routing Problem. Distances between records are calculated by the Google Distance Matrix API. The optimization problem is then solved using Google OR Tools python package.
For more information see <https://developers.google.com/optimization/routing>.  Much of this code is based on the examples provided on that page.

@yvaucher @max3903 @wolfhall @murtuzasaleh @dreispt Could you please review this?

In the meantime, I am working on a PR for the field-service repo that will depend on this framework and provide an example of its use.


wolfhall:
@brian10048 
Thanks for taking this on - this was on my list for a while.  The Google OD matrix and TSP are a good first step. 
Maybe I missed it but do you have a write-up on how you are planning on integrating the TSP?  
Also, are you planning on storing any OD results so they can be re-used for further analysis?  
We used to create a dynamic ODM that we could add to so we don't have to recalculate the entire matrix.  
For small TSPs with just 10-20 stops it is likely not needed.  But when we get into VRP (multiple route assignment and scheduling,) we will need this.

Soem suggestions - maybe for version 2:

It may make sense to split the OD and TSP into two modules.  This would allow us to handle OD calculation independently and also incorporate other algorithms.  Typically we would want a matrix for distance and travel time.  The Google matrix gets us both.

I believe Google also allows to set a maximum cut=off distance to restrict matrix pairs that are outside of a feasible travel distance.  I envision these configuration parameters in admin settings in Odoo.

For the TSP people usually like shortest path (based on distance) and shortest time (based on travel time).  Again these can be parameters.

I did not see the option to set time windows (for appointment time ranges) on the TSP.  Google exposes this on the VRP.  So for future modules we can use VRP for one vehicle to calculate TSP with Time Windows.

I will also make some comments on the FSM module separately.

Wolfgang

base_geo_route_optimization/models/geo_route_optimize.py
```diff
@@ -0,0 +1,189 @@
...
+        ''' Build and send the API request for the given origin and
+           destination addresses.
+        '''
+        def build_address_str(addresses):
+            # Build a pipe-separated string of addresses
+            address_str = ''
+            for i in range(len(addresses) - 1):
+                address_str += addresses[i] + '|'
+            address_str += addresses[-1]
+            return address_str
```
yvaucher: You might just want to use `join` here. Or am I missing something?

```suggestion
            return '|'.join(addresses)
```

base_geo_route_optimization/models/geo_route_optimize.py
```diff
@@ -0,0 +1,189 @@
...
+from ortools.constraint_solver import pywrapcp
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
+
+
+class GeoRouteOptimize(models.AbstractModel):
+    ''' This model is meant to solve the Traveling Salesman Problem
```
yvaucher: Please apply `black` https://github.com/psf/black code formatting on this new module.

As it is now the required code formatting in version 13.0 it will avoid too many diff between versions.

The idea behind `black` is to automate the code formatting and loose no time on such things.

base_geo_route_optimization/models/geo_route_optimize.py
```diff
@@ -0,0 +1,189 @@
...
+           format as required in the prepare_destinations function.
+           Returns a dictionary with a list of the record ids sorted in
+           the optimized order and a list of durations between nodes
+        '''
+        if not origin:
+            raise ValidationError(_('No origin defined for route optimization'))
+        unsorted_ids = self.ids
+        num_vehicles = 1
+        # A depot (start & end point) is the first index in `[addresses]`
+        depot = 0
```
yvaucher: Shouldn't it be a constant?

It seems it's value is always `0`

yvaucher (changes requested):
Only a couple of things and we can do an optimistic merge as it's development_status is Beta.

About black I think it's better to do it now than later to avoid unnecessary noises in the commit history in version 13.0 or later.

base_geo_route_optimization/models/geo_route_optimize.py
brian10048: Good call. Updated

base_geo_route_optimization/models/geo_route_optimize.py
brian10048: Applied black formatting

base_geo_route_optimization/models/geo_route_optimize.py
brian10048: I made it a constant, local to that function.  It shouldn't change

In the future, `DEPOT` may be replaced with a separate start node and end node as the library being used allows this

yvaucher (approved):

brian10048:
> Thanks for taking this on - this was on my list for a while. The Google OD matrix and TSP are a good first step.

Glad to do it @wolfhall! We were spending a lot of time manually building daily schedules from the map view in FSM, so this was something my team needed dearly!

> Maybe I missed it but do you have a write-up on how you are planning on integrating the TSP?

It looks like now you've seen the field-service module I put together that uses a wizard to work through the TSP.  I've also considered creating a cron task that runs on a schedule to build routes in advance automatically.   I tried to keep this as a basic framework, so it can be integrated in a way that suites multiple scenarios.

> Also, are you planning on storing any OD results so they can be re-used for further analysis?
> We used to create a dynamic ODM that we could add to so we don't have to recalculate the entire matrix.
> For small TSPs with just 10-20 stops it is likely not needed. But when we get into VRP (multiple route assignment and scheduling,) we will need this.

I had not thought about storing the results, but I can see how that would be useful.  I haven't thought of how to implement that specifically, but am open to suggestions!
Since this uses a web request to Google for the distance matrix calculations, it shouldn't be that resource intensive on our ends even when calculating for a lot of addresses.  However, using the python library to solve a more complicated VRP problem with these matrix results may slow things down??

> 
> Soem suggestions - maybe for version 2:
> 
> It may make sense to split the OD and TSP into two modules. This would allow us to handle OD calculation independently and also incorporate other algorithms. Typically we would want a matrix for distance and travel time. The Google matrix gets us both.

One API request returns the distance and travel time, so we could store a matrix for each if needed.  

> I believe Google also allows to set a maximum cut=off distance to restrict matrix pairs that are outside of a feasible travel distance. I envision these configuration parameters in admin settings in Odoo.

I know we can set a maximum distance for the overall route. I'm not quite sure how it works for a maximum time between stops...  

> For the TSP people usually like shortest path (based on distance) and shortest time (based on travel time). Again these can be parameters.

I hard-coded this to use travel time (duration).  Company-wide parameters are definitely an option

> I did not see the option to set time windows (for appointment time ranges) on the TSP. Google exposes this on the VRP. So for future modules we can use VRP for one vehicle to calculate TSP with Time Windows.

It's definitely part of the roadmap to incorporate time windows.  It shouldn't be a difficult task to add that to this module, as the python library does all of the legwork, but I think we will need to add some additional fields to store time windows on the FSM orders or FSM locations in field-service


> I will also make some comments on the FSM module separately.
> 
> Wolfgang

brian10048:
Travis lint is red, but the black formatting causes it

base_geo_route_optimization/models/geo_route_optimize.py
```diff
@@ -0,0 +1,190 @@
...
+        except urllib.error.URLError as e:
+            raise ValidationError(
+                _("Failure when making Distance Matrix API request: %s", e)
+            )
+        return response
+
+    def build_distance_matrix(self, response):
+        """The following function builds rows of the distance matrix,
+           using the response returned by the send_request function."""
+        distance_matrix = []
```
AyoubZahid, line 68: I think you must check the status of the response before building the matrix
may be somthing like :
`        if response["status"] != 'OK':
            raise ValidationError(_("Service Api responded with : %s") % response["status"])`

PR #265, opened 2020-07-03
10.0 add choice model on layers

lmarion-source:
[IMP] Add the possibility to use different models on one geoengine map

### DIFF
--- a/base_geoengine/static/src/js/views/geoengine_view.js
+++ b/base_geoengine/static/src/js/views/geoengine_view.js
@@ -266,6 +266,10 @@ var GeoengineView = View.extend(geoengine_common.GeoengineMixin, {
 
     createVectorLayer: function(cfg, data) {
         var self = this;
+        var lv = new ol.layer.Vector({
+            title: cfg.name,
+            active_on_startup: cfg.active_on_startup,
+        });
         if (!!cfg.model){
             var vectorSource = new ol.source.Vector({
                 title: cfg.name,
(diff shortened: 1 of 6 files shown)


lmignon:
@lmarion-source Can you take the 10.0 branch as base and only add your last commit to create your PR. It seems that your branch is not bases on the last version of 10.0

base_geoengine/static/src/js/views/geoengine_view.js
```diff
@@ -229,39 +232,62 @@ var GeoengineView = View.extend(geoengine_common.GeoengineMixin, {
...
+        })
+
+    },
+
+    createVectorLayer: function(cfg, data) {
+        var self = this;
+        if (!!cfg.model_id){
+
+            var model = new Model(cfg.model_id[2]);
+            var fields_to_read = [cfg.geo_field_id[1]];
```
lmignon: @lmarion-source In odoo, the field display_name is present by default on all the models. You should add this field by defaults into the list of retrieved fields and use this field as title of the details form dialog..

```js
var fields_to_read = [cfg.geo_field_id[1], 'display_name'];
```

base_geoengine/static/src/js/views/geoengine_view.js
```diff
@@ -788,18 +814,37 @@ var GeoengineView = View.extend(geoengine_common.GeoengineMixin, {
...
+            }
+        } 
+        else {
+            var model_obj = new Model(model);
+            model_obj.call('get_formview_id', [[oid], self.dataset.context]).then(function(view_id){
+                var pop = new common.FormViewDialog(self, {
+                    res_model: model,
+                    res_id: oid,
+                    context: self.dataset.context,
+                    title: _t("Open: ") + attributes.name,
```
lmignon: @lmarion-source the `name` attribute is not present by default. As proposed above, you should use `display_name`.

lmignon (changes requested):
@lmarion-source Greet feature!  some little comments...

lmignon:
@lmarion-source The name of the fields into the popup are not right when displaying the info of the layer defined for an other model. Into the fields definition given to the `formatFeatureHTML` is the one of the main model not the one of the layre's model.

PR #268, opened 2020-07-24
[FIX] web_view_google_map: JS Files

osi-scampbell:
We need to include these files if we want to be able to use the google address autocomplete widget. If these are included elsewhere please let me know as we have a few modules that depend on this module and are missing these .js files


bodedra (approved):
LGTM :+1:

osimallen (approved):
Looks good

osi-scampbell:
moving to #270

PR #270, opened 2020-07-27
[IMP] web_view_google_map: Separate API keys for Geocode/Autocomplete

osi-scampbell:
The Google Address Autocomplete widget uses different security protocols than the Geocoding Api calls. This PR allows you to use 2 keys however it is still in draft because ideally we only want to have 1 key.

### DIFF
--- a/web_view_google_map/static/src/js/widgets/gplaces_autocomplete.js
+++ b/web_view_google_map/static/src/js/widgets/gplaces_autocomplete.js
@@ -90,8 +90,8 @@ odoo.define('web_view_google_map.GplaceAutocompleteFields', function (require) {
                         center: geolocation,
                         radius: position.coords.accuracy
                     });
-
-                    self.places_autocomplete.setBounds(circle.getBounds());
+                    // Temporary Comment
+                    // self.places_autocomplete.setBounds(circle.getBounds());
                 });
             }
         },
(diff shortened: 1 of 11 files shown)


max3903:
@osi-scampbell Is this ready for review?

osi-scampbell:
@max3903 This PR is good for Pavlov's needs but I don't think it is the final answer for OCA. Ideally I want this module to check if the fields are set before it overwrites certain fields. I have reached out to @dreispt on this and also sent an email to OCA/web. I also sent an email to Yopi since he originally wrote these modules. This is should be part of the PVM release because it satisfies the ticket. If we want to merge this now and then add the additional workflow later I am fine with that too

max3903:
@osi-scampbell Can you add the hook scripts to move the possibly existing value to the new key?

It should fix the error below:

```
odoo.tools.convert.ParseError: "duplicate key value violates unique constraint "ir_config_parameter_key_uniq"
DETAIL:  Key (key)=(google.api_key_geocode) already exists.
```

osi-scampbell:
> @osi-scampbell Can you add the hook scripts to move the possibly existing value to the new key?
> 
> It should fix the error below:
> 
> ```
> odoo.tools.convert.ParseError: "duplicate key value violates unique constraint "ir_config_parameter_key_uniq"
> DETAIL:  Key (key)=(google.api_key_geocode) already exists.
> ```

@max3903 Hooks added

web_view_google_map/static/src/js/widgets/gplaces_autocomplete.js
```diff
@@ -329,7 +329,7 @@ odoo.define('web_view_google_map.GplaceAutocompleteFields', function (require) {
                 general: {
                     name: 'name',
                     website: 'website',
-                    phone: ['international_phone_number', 'formatted_phone_number']
+                    // phone: ['international_phone_number', 'formatted_phone_number']
```
gityopie, line 332: Hi @osi-scampbell 

Instead of add comment to this line (disable `phone`), maybe better to override function `_onUpdateWidgetFields` and prevent the widget overwrite existing value if any
the override function will looks like this
```javascript
_onUpdateWidgetFields: function (values) {
    values = values || {};
    if (Object.prototype.hasOwnProperty.call(this.fillfields.general || '', 'phone')) {
        if (this.recordData.phone) {
            delete values['phone'];
        }
    }
    if (Object.prototype.hasOwnProperty.call(this.fillfields.general || '', 'website')) {
        if (this.recordData.website) {
            delete values['website'];
        }
    }
    this._super(values);
}
```
The override function above will be applied to `website` as well

dreispt:
@osi-scampbell Travis is red and PR is marked as draft. Is it expected to have progress on this?

PR #273, opened 2020-10-26
[12.0][MIG] geoengine_base_geolocalize

ecino:


lmignon (changes requested):
@ecino you should take a look to the 10.0 branch and includes changes from https://github.com/OCA/geospatial/pull/155

PR #276, opened 2020-12-01
[MIG] base_geolocalize_openstreetmap: Migration to 12.0

len-foss:


marcelsavegnago:
@len-foss 
image

len-foss:
Hello @marcelsavegnago thanks for testing!
I don't think it's an issue: `geoengine.raster.layer` is defined in `base_geoengine`, which is not in the dependency tree of `base_geolocalize_openstreetmap` (its only dependency is `base_geolocalize`, in standard).
If you think there's a problem with the security rules you should probably open an issue related to that module.

marcelsavegnago:
> Hello @marcelsavegnago thanks for testing!
> I don't think it's an issue: `geoengine.raster.layer` is defined in `base_geoengine`, which is not in the dependency tree of `base_geolocalize_openstreetmap` (its only dependency is `base_geolocalize`, in standard).
> If you think there's a problem with the security rules you should probably open an issue related to that module.

Hello @len-foss. Thanks for the feedback.

marcelsavegnago (approved):
LGTM

lmignon (approved):

PR #280, opened 2021-01-06
[IMP] base_geoengine: make geoengine rights implied by standard usage…

len-foss:
… groups

Before, installing base_geoengine makes all users except root_user to not
have access to contacts because they do not have access to geoengine models.
Users should not need new rights to perform old activities.
Therefore we make base.group_user imply user rights.
For ease of use, the Geoengine admin group is made implied by group_system.

### DIFF
--- a/base_geoengine/security/data.xml
+++ b/base_geoengine/security/data.xml
@@ -11,4 +11,18 @@
         <field name="users" eval="[(4, ref('base.user_root'))]"/>
     </record>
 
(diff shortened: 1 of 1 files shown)


lmignon (approved):

marcelsavegnago (approved):

len-foss:
@marcelsavegnago see, your comment was right :-)

bodedra (approved):
LG :+1:

PR #281, opened 2021-01-19
[WIP][ADD] base_geoengine_jsonify: support of geo_point json exports

len-foss:
Dependent on https://github.com/OCA/server-tools/pull/1894 refactoring to actually work.

### DIFF
--- a/base_geoengine_jsonify/__manifest__.py
+++ b/base_geoengine_jsonify/__manifest__.py
@@ -1,0 +1,14 @@
+# Copyright 2020 ACSONE SA/NV.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Support of GeoPoint Json exports for Odoo",
+    "version": "13.0.1.0.0",
+    "category": "GeoBI",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/geospatial",
+    "depends": ["base_geoengine", "base_jsonify"],
+    "external_dependencies": {"python": ["geojson"]},
+    "installable": True,
+    "auto_install": True,
+}
--- a/base_geoengine_jsonify/models/models.py
+++ b/base_geoengine_jsonify/models/models.py
@@ -1,0 +1,19 @@
+# Copyright 2020 ACSONE SA/NV.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import geojson
+
+from odoo import models
+from  odoo.addons.base_geoengine.fields import GeoField
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _jsonify_value(self, field, value):
+        """Adds support for geo_point fields."""
+        res = super(Base, self)._jsonify_value(field, value)
+        if res and isinstance(field, GeoField):
+            # we eval it as dumps gives a str, we need a point for json.dumps
+            res = geojson.loads(geojson.dumps(res))
+        return res
--- a/base_geoengine_jsonify/readme/DESCRIPTION.rst
+++ b/base_geoengine_jsonify/readme/DESCRIPTION.rst
@@ -1,0 +1,1 @@
+Automatically export geopoint fields as geojson in json exports.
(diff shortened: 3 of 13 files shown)


base_geoengine_jsonify/models/models.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import models, tools
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _jsonify_value(self, field, value):
+        """Adds support for geo_point fields."""
+        res = super(Base, self)._jsonify_value(field, value)
+        if res and field.type == "geo_point":
```
lmignon: @len-foss The same logic could be applied to all the geo fields non? 

```python

from  odoo.addons.base_geoengine.fields import GeoField

....

if res and isinstance(field, GeoField):
    res = tools.safe_eval(geojson.dumps(res))
...

```

lmignon (approved):
Great!
LGTM once OCA/server-tools#1894  is merged

base_geoengine_jsonify/models/models.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _jsonify_value(self, field, value):
+        """Adds support for geo_point fields."""
+        res = super(Base, self)._jsonify_value(field, value)
+        if res and isinstance(field, GeoField):
+            # we eval it as dumps gives a str, we need a point for json.dumps
+            res = tools.safe_eval(geojson.dumps(res))
```
sbidoul: If `geojson.dumps` produces a json str, is it not better to parse it with `json.loads` ?

yvaucher:
OCA/server-tools#1894 was merged on Jan 21

yvaucher (approved):
LGTM

Looks interesting

base_geoengine_jsonify/models/models.py
yvaucher: This seems to have been resolved with:
 
 ``` python
           # we eval it as dumps gives a str, we need a point for json.dumps
            res = geojson.loads(geojson.dumps(res))
```

Seems weird, waiting on travis results (don't know what to expect from the content of `res` in this context)

**Comment by yvaucher** [comment]
Just relaunched the CI

File: base_geoengine_jsonify/models/models.py [hunk 3]
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo import models
+from  odoo.addons.base_geoengine.fields import GeoField
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def _jsonify_value(self, field, value):
+        """Adds support for geo_point fields."""
```
yvaucher, line 14: Let's be a bit more generic, geojson.dumps should support multiline string and polygones too.

```suggestion
        """Adds support for geometry fields."""
```

base_geoengine_jsonify/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Automatically export geopoint fields as geojson in json exports.
```
yvaucher, line 1:
```suggestion
Automatically export geometry fields as geojson in json exports.
```

base_geoengine_jsonify/models/models.py
```diff
@@ -0,0 +1,19 @@
...
+        res = super(Base, self)._jsonify_value(field, value)
+        if res and isinstance(field, GeoField):
+            # we eval it as dumps gives a str, we need a point for json.dumps
```
yvaucher, line 17:
```suggestion
            # we eval it as dumps gives a str, we need a geometry for json.dumps
```

base_geoengine_jsonify/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+Nothing to do!
+If you add a geo_point to a json export, the point will be converted automatically
```
yvaucher, line 2:
```suggestion
If you add a geometry to a json export, the point will be converted automatically
```

base_geoengine_jsonify/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 ACSONE SA/NV.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Support of GeoPoint Json exports for Odoo",
```
yvaucher, line 4:
```suggestion
    "name": "Export geometry as GeoJSON",
```

yvaucher (approved):
Nice, few suggestion to make the module a bit more generic in the terminology as GeoPoint is not the only geometry that can be handled. I assume it would work too, to create a GeoJSON for a Polygon or a LineString.

PR #282, opened 2021-01-26
[FIX] vertex layer label (generalize to all layer types)

StephaneMangin:
Problem encountered:
Labels on vertex layers by feature was not working anymore (assuming since the version 12 or 10)
Despite the fact that a boolean field called display label was still present in the code, this behavior was totally broken.

This issue has been addressed and label display is now available to all custom types of vertex layer's by feature (basic, colored, proportion)


sebalix:
You have to install `pre-commit` starting from 13.0 repositories

PR #287, opened 2021-03-10
[MIG] base_google_map: Migration to 13.0

brian10048:


brian10048:
@yvaucher - There is the same error here as well https://travis-ci.com/github/OCA/geospatial/jobs/489824581#L581

yvaucher:
https://github.com/OCA/geospatial/pull/287/checks?check_run_id=2327638791

The lint is now under github actions, it's just missing some formating of one file and we can merge

(running pre-commit should auto format the file for you)

yvaucher (approved):

brian10048:
Thank you @yvaucher - I ran pre-commit on this module.  Looks like there are some files in other modules that are causing lint to fail

yvaucher:
@brian10048 IIRC correcly the pre-commit on travis was run only on the changes. Thanks for the lint fixing PR. #294

PR #290, opened 2021-04-07
[14.0][MIG] base_google_map

AmmarOfficewalaSerpentCS:


AmmarOfficewalaSerpentCS:
@smangukiya , 
Please review this PR.

AmmarOfficewalaSerpentCS:
@murtuzasaleh , @dreispt, @max3903 , @brian10048  
can you please review this PR?

base_google_map/data/google_maps_libraries.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
patrickrwilson, line 1: This line is no longer needed (since v11)

base_google_map/views/res_config_settings.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
patrickrwilson: This line is no longer needed (since v11)

base_google_map/views/res_config_settings.xml
```diff
@@ -0,0 +1,153 @@
...
+<odoo>
+    <data>
```
patrickrwilson: data tag no longer needed.

base_google_map/data/google_maps_libraries.xml
```diff
@@ -0,0 +1,10 @@
...
+<odoo>
+    <data noupdate="1">
```
patrickrwilson: data tag no longer needed, noupdate should be moved into odoo tag.

patrickrwilson (approved):
A few very small changes requested but no blockers.

AmmarOfficewalaSerpentCS:
@patrickrwilson Can Please Review it.

smangukiya (approved):

smangukiya:
@lmignon @yvaucher Would you mind reviewing it and if it looks good then merge it? Appreciated. Thank you

lmignon:
@dreispt PLZ don't use `nobump` when merging. When you use the `nombup` option, only a 'dev' release is published on pypi..

lmignon (changes requested):
@AmmarOfficewalaSerpentCS @smangukiya Than you for the migration. Can you check and fix the pre-commit error? https://github.com/OCA/geospatial/pull/290/checks?check_run_id=2980089610#step:4:165

AmmarOfficewalaSerpentCS:
@lmignon , 
Pre-commit fix 
Can you please review it again ?

Freni-OSI:
CC: @patrickrwilson

dreispt:
@lmignon Regarding `nobump`, this is not clear to me.
I'm not comfortable forcing a version bump in the cases it is not needed.
I could not find good guidance on this from the docs I found, and even from the bot source code.
@sbidoul Can you confirm this?

PS: This is off-topic, maybe we should move this discussion elsewhere.

lmignon:
@dreispt AFAIK without version change, the addon will be published on pypi as a dev version. As result, the installation will only be possible through pip by using the `--dev` flag. The result of the migration should be a stable version not a dev version. That's why a version bump is required.

sbidoul:
@lmignon @dreispt 

A few things about nobump and round version numbers on PyPI

- When using `ocabot merge patch/minor/major`, the bot bumps version and makes the wheel on the same commit it bumped the version, which leads to a "round" (non-dev) version number. When using nobump or a manual merge, the wheel gets generated on the HEAD of main branch which is not necessarily a commit that bumped the manifest version, which leads to a `.dev` version number (counting the commits since the last manifest bump).
- For migration PRs, I tend to use `patch` to be sure to get a round version number.
- That said, I recently noticed that when there are *only* dev versions, pip picks one even without using `--pre` so it does not actually matter that much in practice.
- It's been a while now that I concluded that these `.dev` versions are not useful for OCA and I reckon that in OCA merging is releasing, and trying to be any smarter than that is never going to work. So for 15.0 I plan to use the new setuptools-odoo versioning mode that adds a 6th digit to the version instead of a `.devN`. So there will no need to worry about all that anymore.

PR #294, opened 2021-04-14
[IMP] pre-commit

brian10048:
@yvaucher Let me know if this is OK to run pre-commit on all the existing modules in one commit

### DIFF
--- a/base_geoengine/geo_ir/ir_model_view.xml
+++ b/base_geoengine/geo_ir/ir_model_view.xml
@@ -1,23 +1,31 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8" ?>
 <odoo>
 
     <record model="ir.ui.view" id="view_model_fields_form_geo">
         <!-- must be unique in this module. -->
         <field name="name">ir.model.fields.form.geo</field>
         <field name="model">ir.model.fields</field>
-        <field name="inherit_id" ref="base.view_model_fields_form"/>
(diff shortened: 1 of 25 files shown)


yvaucher:
pre-commit seems to stil complain. I think we are missing some ignore list on base_geoengine/doc/*

Please change the commit message to `[LINT] pre-commit`, should make it easier to spot as a linting change to be able to use locally `--ignore-revs`.

We could probably add this commit later into a `.git-blame-ignore-revs` file especially if it becomes supported by GitHub at some point. (but no rush on that, just giving a `[LINT]` should be good.

brian10048:
@yvaucher I fixed the base_geoengine/doc/* files and updated commit message.

There is still an error with some JS files.  I can look into the issues soon.  Would you prefer the JS changes to be a separate commit since they weren't necessarily fixed by running pre-commit?  Or do I have an issue locally with my pre-commit?

yvaucher:
No need to fixup the content of `base_geoengine/doc`

I just pushed that commit which should ignore those files: https://github.com/OCA/geospatial/commit/f2c9b01467352c16d0276444bad39c9bd5041cd5

brian10048:
I rebased and removed any changes to base_geoengine/doc files

What do you suggest for the JS files causing the failure now?

yvaucher:
Let them like that, I'll fix that later.

I think those were addressed in the eslint config file at some point I'll have to tweak the lint config a bit.

PR #304, opened 2022-03-09
[12.0][FIX] base_geoengine

Martronic-SA:
Fix base_geoengine so it can display popups


yvaucher:
@Martronic-SA please provide a better self explaining message in the commit messages.

The filename is not necessary

Format it like

```
[FIX] base_geoengine: fix xyz

details on xyz
```

PR #307, opened 2022-06-10
[13.0][FIX] Name and order of parameters of the gist index creation function

sersanchus:
It's been a while since spatial indexes could not be created on geographic fields. This commit correctly uses the nomenclature of the functions and the parameters to use.

### DIFF
--- a/base_geoengine/fields.py
+++ b/base_geoengine/fields.py
@@ -7,7 +7,7 @@ from operator import attrgetter
 from odoo import _, fields
 from odoo.tools import sql
 
-from .geo_db import create_geo_column
+from .geo_db import create_geo_column, create_geo_index
 from .geo_helper import geo_convertion_helper as convert
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


yvaucher (approved):
Indeed, good catch, this comes from my commit https://github.com/OCA/geospatial/commit/d4e82d744c59529f9ba7c39dff258b6501368574 where I moved DB related operation in a separate file.

sersanchus:
There are some coverage tests that fail but have nothing to do with this fix. The merge must be accepted anyway.

yvaucher:
@sersanchus The CI should have been fixed by #325 , can you rebase?

sersanchus:
> @sersanchus The CI should have been fixed by #325 , can you rebase?

Hi @yvaucher, already rebased from OCA:13.0 and pushed to my branch.

yvaucher:
Please use `git rebase` instead of `git pull`

sersanchus:
> Please use `git rebase` instead of `git pull`

Ok, my fault... I've rebased now.

yvaucher:
Thanks for the rebase, let's try again.

PR #308, opened 2022-06-10
[12.0][FIX] Name and order of parameters of the gist index creation function

sersanchus:
It's been a while since spatial indexes could not be created on geographic fields. This commit correctly uses the nomenclature of the functions and the parameters to use.

### DIFF
--- a/base_geoengine/fields.py
+++ b/base_geoengine/fields.py
@@ -8,7 +8,7 @@ from odoo import fields, _
 from odoo.tools import sql
 
 from .geo_helper import geo_convertion_helper as convert
-from .geo_db import create_geo_column
+from .geo_db import create_geo_column, create_geo_index
 
 logger = logging.getLogger(__name__)
 try:
@@ -145,7 +145,7 @@ class GeoField(fields.Field):
             index = cr.fetchone()
             if index:
                 return True
-            self._create_index(cr, model._table, self.name)
+            create_geo_index(cr, model._table, self.name)
         return True
 
     def update_db_column(self, model, column):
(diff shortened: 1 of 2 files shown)


yvaucher (approved):
Indeed, good catch, this comes from my commit https://github.com/OCA/geospatial/commit/d4e82d744c59529f9ba7c39dff258b6501368574 where I moved DB related operation in a separate file.

sersanchus:
There are some coverage tests that fail but have nothing to do with this fix. The merge must be accepted anyway.

yvaucher:
The coverage shouldn't block the merge

yvaucher:
The error was 
```
----------

FAIL

FAIL: test_geo_localize (odoo.addons.geoengine_base_geolocalize.tests.test_geoengine_partner.TestGeoenginePartner)

Traceback (most recent call last):

`   File "/home/travis/build/OCA/geospatial/geoengine_base_geolocalize/tests/test_geoengine_partner.py", line 35, in test_geo_localize

`     "Latitude Should be equals",

` AssertionError: 49.95506 != 49.95353 within 5 places : Latitude Should be equals

FAILED

Module geoengine_base_geolocalize: 1 failures, 0 errors

At least one test failed when loading the modules.

Message not found: 'Modules loaded.'
```

I doubt the index should have any impact. I relaunched the merge to check if the issue persists.

sersanchus:
That is the coverage test problem I commented before, the call to 'create_geo_index' is not covered by any unit test...

Check warning on line 148 in base_geoengine/fields.py

Codecov
/ codecov/patch
base_geoengine/fields.py#L148
Added line #L148 was not covered by tests

yvaucher:
@sersanchus The code coverage should not be a blocking check, it seems the CI failed on the tests with the ocabot:
https://github.com/OCA/geospatial/runs/8135928245

yvaucher:
Trying a new merge now that we use github actions

PR #313, opened 2022-09-22
[MIG] base_geoengine 15

shurshilov:
[MIG] base_geoengine 15


NL66278:
@shurshilov Please use the migration method that keeps the commit history: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

NL66278 (changes requested):
Need to follow the standard migration method for OCA modules. Also note that the asShape method has been deprecated and must be replaced.

yvaucher:
Warning the v14 was not migrated properly, thus we have to redo it from v13 and take additional commits from https://github.com/OCA/geospatial/pull/316

yvaucher (changes requested):
Missing git history

yvaucher:
Please refer to #326

Issue #318, opened 2022-12-30
NameError: name 'asShape' is not defined


halie412:
<!-- Provide a general summary of the issue in the Title above.
In version 2.0.0 launched on 12/12/2022 asShape was deprecated

## Module
base_geoengine

## Describe the bug
base_geoengine/geo_helper/geo_convertion_helper.py line 24 

## To Reproduce
**Affected versions**:
I test on version 14

**Additional context**
Solved this using last stable version of Shapy 1 (1.8.5) with

pip uninstall shapely
pip install shapely==1.8.5

**Note**
I posted this issue because of would be usseful for someone, please mark as solved

halie412:
On version 2.0 of 'shapely' was deprecated 'asShape'

this was solved whith:
pip uninstall shapely
pip install shapely==1.8.5

mathben:
https://github.com/shapely/shapely/pull/961/files

```
   .. deprecated:: 1.8
       The proxy geometries (adapter classes) created by this function are
       deprecated, and this function will be removed in Shapely 2.0.
       Use the `shape` function instead to convert a GeoJSON-like dict
       to a Shapely geometry.
```

mathben:
Check #321

PR #320, opened 2023-01-09
[14.0][FIX]base_geoengine: require Shapely<2

HekkiMelody:
#318 

Module will require refactoring to function with Shapely v2. For now, I've added documentation, instructions and a notice in the try/except around the import.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -51,7 +51,7 @@ repo_description: 'This project will enable real life GIS support on Odoo.
 
     * System: *  **PostGIS** http://postgis.refractions.net/
 
-    * Python: *  **Shapely** http://pypi.python.org/pypi/Shapely
+    * Python: *  **Shapely** http://pypi.python.org/pypi/Shapely  (not yet compatible with v2.0)
 
     *  **geojson** http://pypi.python.org/pypi/geojson'
 repo_name: Geospatial Addons for Odoo
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ A geospatial API is provided, to add your own functionalites at your convenience
 
 * System: *  **PostGIS** http://postgis.refractions.net/
 
-* Python: *  **Shapely** http://pypi.python.org/pypi/Shapely
+* Python: *  **Shapely** http://pypi.python.org/pypi/Shapely (not yet compatible with v2.0)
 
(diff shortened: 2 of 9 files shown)


NL66278:
@aleuffre Would any more changes required then replacing `asShape` with `shape`?

HekkiMelody:
@NL66278 I do not know. The upgrade documentation mentions a bunch of things to potentially fix https://shapely.readthedocs.io/en/stable/migration.html , and I honestly didn't have time to look into it. Happy to close my PR in favour of another one that actually makes the module compatible with Shapely v2.

NL66278:
@aleuffre As all tests run fine with replacing asShape with shape, I think this will be enough. I did not look at all the behavioral changes and what consequences that might have, but none of the other removed functions or attributes mentioned in the migration guide are used in 2.0.0.

HekkiMelody:
Was being conservative since I didn't have much time to look into the issue, but sounds good to me. Closing in favour of #322 

Thank you @NL66278

yvaucher:
Shapely break change seems quite recent, released on Dec 12, 2022, we could be nice an give a dedicated error message.
On the other hand shape is already available since shapely 1.8.0 released on Oct 22, 2021.

Sadly, external-dependencies in `__manifest__.py` doesn't allow a check on a minimal version.

NL66278:
Actually shape has been available since 2008:
```
commit 7610854e7344d27c082f4c5f976d661380d1ebc3
Author: Sean Gillies <sean.gillies@gmail.com>
Date:   Mon Apr 28 03:01:30 2008 +0000

    Most users should use shape() instead of asShape()
    
    git-svn-id: http://svn.gispython.org/svn/gispy/Shapely/trunk@1087 b426a367-1105-0410-b9ff-cdf4ab011145

diff --git a/CHANGES.txt b/CHANGES.txt
index a673b16..86c0dc9 100644
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ All tickets are children of http://trac.gispython.org/projects/PCL/ticket/
 - Disentangle Python and topological equality (#163).
 - Implement safe predicate shortcuts for objects identical in the Python sense
   (#163).
+- Add shape(), a factory that copies coordinates from a geo interface provider.
 
 1.0.3 (2008-04-09)
 ------------------
```

PR #321, opened 2023-01-09
[12.0][FIX] base_geoengine. use shape() instead of asShape()

NL66278:
The method asShape has been removed from shapely.geometry and should be replaced by shape.

### DIFF
--- a/base_geoengine/fields.py
+++ b/base_geoengine/fields.py
@@ -12,7 +12,7 @@ from .geo_db import create_geo_column
 
 logger = logging.getLogger(__name__)
 try:
-    from shapely.geometry import asShape
+    from shapely.geometry import shape
     from shapely.geometry import Point
     from shapely.geometry.base import BaseGeometry
     from shapely.wkb import loads as wkbloads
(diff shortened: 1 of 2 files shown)


NL66278:
@lmignon Laurent would you have a look at this please?

lmignon (changes requested):
@NL66278 Thank you for the fix. In the same time, you should also modify https://github.com/OCA/geospatial/blob/12.0/setup/base_geoengine/setup.py to add a dependency override to specify a constraint on the expected version for the shapely lib...

```python
import setuptools

setuptools.setup(
    setup_requires=['setuptools-odoo'],
    odoo_addon={
        'external_dependencies_override': {
            'python': {
                'shapely': 'shapely>=?.?.?'
            },
        },
    }
)
```

NL66278:
@lmignon Is that really needed, as the alternative `shape()` function has been available since 2008, so this should work on both older and newer versions.

lmignon (approved):

mathben (approved):
I tested with my other module, all works, thank you!

PR #324, opened 2023-01-10
[14.0][FIX]base_geoengine:permission error for non-admins

HekkiMelody:
Adding `sudo()` to this search, as otherwise no user can ever see a map unless they have admin rights.


yvaucher (changes requested):
@aleuffre please check your access rights on ir.ui.view and other geoengine related models (raster layer, vector layer).

Otherwise, it might be some shenanigans in other models where a read access is missing.
This sudo seems to be a bad idea to me.

HekkiMelody:
> @aleuffre please check your access rights on ir.ui.view and other geoengine related models (raster layer, vector layer).
> 
> Otherwise, it might be some shenanigans in other models where a read access is missing. This sudo seems to be a bad idea to me.

@yvaucher Thanks for the feedback! Unfortunately, I'm not quite sure I understood what you mean.

In odoo v12 and v13 the line in question worked because `base.group_user` could read `ir.ui.view` but this was removed in v14 https://github.com/odoo/odoo/commit/ccc98e01691b88c2da9c484c0e6de235b9d18548

However, people without administrative rights still need to be able to see the map.

Similar uses of `sudo()` can be found throughout the base `ir.ui.view` model in similar circumstances. https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_ui_view.py

However, maybe I just misunderstood. How do you suggest avoiding `sudo()` in this instance? I'd be more than happy to not use it (even if it's just for one `search()`) and would also feel safer doing so.

base_geoengine/geo_model.py
```diff
@@ -41,8 +41,10 @@ def fields_get(self, allfields=None, attributes=None):
 
     @api.model
     def _get_geo_view(self):
-        geo_view = self.env["ir.ui.view"].search(
-            [("model", "=", self._name), ("type", "=", "geoengine")], limit=1
+        geo_view = (
```
NL66278: Why not:
```
geo_view = self.sudo().env["ir.ui.view"].search(
    [("model", "=", self._name), ("type", "=", "geoengine")],
    limit=1
)
```

NL66278: Two line less code and no unneeded brackets around the assignment expression.

HekkiMelody: Because Black won't let me :cry:

NL66278: @aleuffre I see, this works though:
```
        view_model = self.env["ir.ui.view"]
        geo_view = view_model.sudo().search(
            [("model", "=", self._name), ("type", "=", "geoengine")],
            limit=1,
        )
```

base_geoengine/geo_model.py
```diff
@@ -41,8 +41,10 @@ def fields_get(self, allfields=None, attributes=None):
...
+            self.sudo()
+            .env["ir.ui.view"]
+            .search([("model", "=", self._name), ("type", "=", "geoengine")], limit=1)
         )
         if not geo_view:
             raise except_orm(
```
NL66278: except_orm is deprecated. Replace by UserError.

HekkiMelody: I don't mind including the change, but I feel like that should be a different PR.

HekkiMelody:
Rebased on https://github.com/OCA/geospatial/pull/326

yvaucher:
ok, now I see it:
https://github.com/odoo/odoo/blob/14.0/odoo/models.py#L1542

Thanks for the fix. I'll add it directly into #326 to save you some rebase.

yvaucher:
(thus closing it as this commit will be part of the (re-) migration )

PR #328, opened 2023-02-14
[16.0] [MIG] base_geoengine : migration from odoo 14 to odoo 16

kouffsamuel:


yvaucher:
@lmignon FYI, BTW does acsone still plan to migrate geoengine on Owl ?

yvaucher:
Commit to add :

https://github.com/OCA/geospatial/pull/327

lmignon:
Hi @yvaucher 

@kouffsamuel will migrate the widgets and views to owl. He is working under my supervision on this subject as part of his final thesis (supervision is perhaps not the most appropriate word, we can say that we are collaborating even if he is doing all the work and I am there to guide and advise him in what he is doing). I proposed to him to open a draft PR in order to give visibility to his work which will probably take several weeks.  He will make a second PR with the demo module so that external people can test functionally what will be done. Any comment, suggestion for evolution etc. will be welcome.  

Have a nice day.

base_geoengine/static/src/js/widgets/geoengine_edit_map/field_geoengine_edit_map.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <templates xml:space="preserve">
     <t t-name="FieldGeoEngineEditMap" owl="1">
-
+        <div id="map" style="width: 100%; height: 100%;">
```
lmignon: Do we need a `style` attribute on this element? You already specify a css rule on 'o_field_geo_edit_map'. In any case, if style attributes are needed, they must come from the css and not from the style attribute on the element itself.

The value for the `id` attribute must come from the widget id to avoid duplicate if the widget is used more than one time on the same view.

lmignon: ping @kouffsamuel

dangeross:
Hi @kouffsamuel 

Seems like [base_geoengine/static/lib/geostats-2.0.0](https://github.com/OCA/geospatial/pull/328/commits/ce6e046c80162f3c69f439d3acb5a3632214f1ef#diff-33c2bcd08018ad8e4da200f5b91f5b34a9134a0f52b50f9fda9f8d6245e6dc46) is referencing a submodule somewhere (by mistake?)

base_geoengine/fields.py
```diff
@@ -0,0 +1,290 @@
...
+
+        value can be geojson, wkt, shapely geometry object.
+        If geo_direct_write in context you can pass diretly WKT"""
+        if not value:
+            return None
+        shape_to_write = self.entry_to_shape(value, same_type=True)
+        if shape_to_write.is_empty:
+            return None
+        else:
+            return shape_to_write.wkt
```
sbejaoui: @lmignon , @kouffsamuel

I am getting an error when editing a geoField,

```sql
UPDATE "res_partner" SET "geo_point" = 'POINT (-13578158.298793 4513301.6464)';
```
```log
psycopg2.errors.InvalidParameterValue: Geometry SRID (0) does not match column SRID (3857)
```
I fixed it by setting the SRID for the given point 

 ```suggestion
            return f'SRID={self.srid};{shape_to_write.wkt}'
```

File: base_geoengine/fields.py [hunk 3]
```diff
@@ -0,0 +1,290 @@
...
+            SELECT
+                ST_Transform(
+                    ST_GeomFromText(%(wkt)s, 4326),
+                    %(srid)s)
+        """,
+            {"wkt": pt.wkt, "srid": cls.srid},
+        )
+        res = cr.fetchone()
+        return cls.load_geo(res[0])
+
```
lmignon, line 255: @kouffsamuel This commit is missing https://github.com/OCA/geospatial/commit/f5dfefb7d74a40c1c2684a019c7f1f609b6d8c26

base_geoengine/fields.py
lmignon: @kouffsamuel @sbejaoui This conversion was previously implemented in 10.0 by the use of the 'column_format' on the field definition https://github.com/OCA/geospatial/blob/10.0/base_geoengine/fields.py#L36 
```python
    @property
    def column_format(self):
        return "ST_GeomFromText(%%s, %s)" % (self.srid,)
```

The `convert_to_column` method seems to be the replacement of the `column_format` method.  Would it be more accurate to write?
```suggestion
            return f"ST_GeomFromText({shape_to_write.wkt}, {self.srid})"
```

base_geoengine/expressions.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+        case "geo_contains":
+            query = current_operator.get_geo_contains_sql(
+                table, left, right, params, rel_col=rel_col, rel_model=rel_model
+            )
+        case "geo_intersect":
+            query = current_operator.get_geo_intersect_sql(
+                table, left, right, params, rel_col=rel_col, rel_model=rel_model
+            )
+    return query
```
sbejaoui, line 135:
```suggestion
        case _:
            raise NotImplementedError(f'The operator {operator} is not supported')
    return query
```

base_geoengine/models/base.py
```diff
@@ -0,0 +1,169 @@
...
+         * geo_within
+         * geo_contains
+         * geo_intersect"""
+        # First we do a standard search in order to apply security rules
+        # and do a search on standard attributes
+        # Limit and offset are managed after, we may loose a lot of performance
+        # here
+        _logger.debug(
+            "geo_search is deprecated: uses search method defined on base model"
+        )
```
lmignon, line 157: @kouffsamuel We can use here the standard way to declare deprecation into python....
 In your imports add: `import warnings`

```suggestion
        warnings.warn(
           "geo_search is deprecated: uses search method defined on base model",
        DeprecationWarning)
```

lmignon (approved):
Great work @kouffsamuel ! LGTM

drewes:
Fantastic work. Well done.

baimont:
A general comment, which is quite subjective:

image

You commit a lot. This makes it quite difficult to read the great categories of changes that you have done during the migration. But as I said, this is quite subjective for I myself do not commit enough and tend to make one ugle big commit for a migration, even if it's that big. Nevertheless, maybe some of these commits could be merged together?

Anyway, fantastic work! Well done!

base_geoengine/readme/USAGE.rst
```diff
@@ -0,0 +1,43 @@
+--------------
+Geoengine Demo
+--------------
+1. As a user/admin, when I am in the Geoengine Demo module and I go to the ZIP menu.
+   When I click on an item in the list view, I get to the form view showing me the different
+   information about the ZIP. We can see its ZIP, city, priority, total sales and his spatial
```
baimont:
```suggestion
   information about the ZIP. We can see its ZIP, city, priority, total sales and its spatial
```

kouffsamuel:
> A general comment, which is quite subjective:
> 
> image
> 
> You commit a lot. This makes it quite difficult to read the great categories of changes that you have done during the migration. But as I said, this is quite subjective for I myself do not commit enough and tend to make one ugle big commit for a migration, even if it's that big. Nevertheless, maybe some of these commits could be merged together?
> 
> Anyway, fantastic work! Well done!

Thanks for your feedback.

yvaucher:
OL 7 :yum: 

Fantastic work.

yvaucher (commented):
On the UI, the GeoEngine Data tab on ir.ui.view form has some fields (projection and restricted map extent `restricted_extent`) that are not well displayed.

~~Do you also plan to migrate base_geoengine_demo?~~ (https://github.com/OCA/geospatial/pull/329)

yvaucher:
@kouffsamuel If I understand correctly,  test_base_geoengine is integrated in `base_geoengine` as I see some new test files?

Or should we still migrate it?

CF: https://github.com/OCA/geospatial/blob/13.0/test_base_geoengine/tests/test_base_geoengine.py

kouffsamuel:
> @kouffsamuel If I understand correctly, test_base_geoengine is integrated in `base_geoengine` as I see some new test files?
> 
> Or should we still migrate it?
> 
> CF: https://github.com/OCA/geospatial/blob/13.0/test_base_geoengine/tests/test_base_geoengine.py

Yes, I do some tests in base_geoengine but I didn't take into account these tests. So it should probably still be migrated.

lmignon:
> > @kouffsamuel If I understand correctly, test_base_geoengine is integrated in `base_geoengine` as I see some new test files?
> > Or should we still migrate it?
> > CF: https://github.com/OCA/geospatial/blob/13.0/test_base_geoengine/tests/test_base_geoengine.py
> 
> Yes, I do some tests in base_geoengine but I didn't take into account these tests. So it should probably still be migrated.

IMO, the test_base_geoengine addon is now useless. Tests in base_geoengine covers more cases.

lmignon:
@yvaucher Do you agree to merge this one... I would like to move forward with all the pending PR's. If we encouter bugs we'll fix it since these addons are used in our current project.

bealdav:
Hi @yvaucher could you actualize your review, please. Thanks

yvaucher (approved):

lmignon:
merged with #334

PR #334, opened 2023-03-20
[16.0] [ADD] base_geoengine: editing vector layers on geoengine view

kouffsamuel:
* [ ] depends on #328 
* [ ] depends on #329 

This PR will add an edit button next to vector layers in the geoengine view to edit the layer that we want to used.

PR #338, opened 2023-04-28
[16.0][ADD] base_geoengine: geoengine view in edit mode

kouffsamuel:
This PR will allow users to modify or create new records from the geoengine view.

* [ ] depends on #334


lmignon (approved):

lmignon:
@ged-odoo It would be nice to have your comments on this work in which @kouffsamuel has tried to apply patterns understood through reverse engineering of Odoo components to develop new widgets and an advanced view for handling GIS data. 
All this work can be tested via https://runboat.odoo-community.org/webui/build.html?name=b3bb0751e-240d-46f8-96e5-a77697ac615b

lmignon:
@kouffsamuel Can you add a test-requirements.txt to include base_geoengine_demo into the build so it will be possible to tests this branch from the runboat instance linked to the PR. The one on `base_geoengine_demo` doesn't take into account your last change since it has been build on the last put of the `base_geoengine_demo`  migration branch.

kouffsamuel:
> @kouffsamuel Can you add a test-requirements.txt to include base_geoengine_demo into the build so it will be possible to tests this branch from the runboat instance linked to the PR. The one on `base_geoengine_demo` doesn't take into account your last change since it has been build on the last put of the `base_geoengine_demo` migration branch.

Like this ? https://github.com/OCA/geospatial/pull/338/commits/3ca04d85ab017db473ecff507f28d25a82bde0f9

base_geoengine/static/src/js/vector_layers_store.esm.js
```diff
@@ -0,0 +1,36 @@
+/** @odoo-module */
+const {reactive} = owl;
```
ged-odoo: random remark: you can import owl from `@odoo/owl`, and it is now the recommended way to do it.
```js
import { reactive } from "@odoo/owl"
```

base_geoengine/static/src/css/style.css
```diff
@@ -0,0 +1,402 @@
+/* OpenLayers Style */
```
ged-odoo, line 1: in odoo, we decided to remove the `/css`, `/js` and `/xml` subfolders. The new guideline is to put everything in `/static/src`, and locate related code together (so, typically, the xml/js/scss for a component should be located at the same place

base_geoengine/static/src/js/raster_layers_store.esm.js
```diff
@@ -0,0 +1,41 @@
...
+
+    getRaster(id) {
+        return this.rasters.find((el) => el.id === id);
+    }
+
+    get count() {
+        return this.rasters.length;
+    }
+}
+export const rasterLayersStore = reactive(new RasterLayersStore());
```
ged-odoo, line 42: looks good to me. The `reactive` function opened up a lot of new ways of organizing the code.

base_geoengine/static/src/js/views/geoengine/geoengine_arch_parser.esm.js
```diff
@@ -0,0 +1,86 @@
...
+     */
+    parse(arch, models, modelName) {
+        const xmlDoc = this.parseXML(arch);
+        const templateDocs = {};
+        const fieldNodes = {};
+        const jsClass = xmlDoc.getAttribute("js_class");
+        const activeFields = {};
+        const geoengineAttr = {};
+
+        this.visitXML(xmlDoc, (node) => {
```
ged-odoo, line 31: i am not a fan of our archparser system. it feels way too complicated. sorry for that.

base_geoengine/static/src/js/views/geoengine/layers_panel/layers_panel.esm.js
```diff
@@ -0,0 +1,241 @@
...
+import {useOwnedDialogs, useService} from "@web/core/utils/hooks";
+import {DomainSelectorGeoFieldDialog} from "../../../widgets/domain_selector_geo_field/domain_selector_geo_field_dialog/domain_selector_geo_field_dialog.esm";
+import {FormViewDialog} from "@web/views/view_dialogs/form_view_dialog";
+import {useSortable} from "@web/core/utils/sortable";
+
+const {Component, onWillStart, useState, useRef} = owl;
+
+export class LayersPanel extends Component {
+    setup() {
+        super.setup();
```
ged-odoo: you don't have to call super.setup if you inherit from Component. Owl guarantees that the setup function of Component is always empty

base_geoengine/static/src/js/views/geoengine/layers_panel/layers_panel.esm.js
```diff
@@ -0,0 +1,241 @@
...
+        this.state = useState({geoengineLayers: {}, isFolded: false});
+        this.addDialog = useOwnedDialogs();
+        let dataRowId = "";
+
+        /**
+         * Call the model method "get_geoengine_layers" to get all the layers
+         * in the database and add them to the store.
+         */
+        onWillStart(async () => {
+            this.isGeoengineAdmin = await this.user.hasGroup(
```
ged-odoo: be aware that the 2 calls here will be made sequentially. You can change the code by using Promise.all to make them in parallel (it's not a big deal however, the user group check will only be made once, but still)

base_geoengine/static/src/js/views/geoengine/layers_panel/layers_panel.esm.js
```diff
@@ -0,0 +1,241 @@
...
+    }
+    /**
+     * This method allows you to open/close the panel.
+     */
+    fold() {
+        this.state.isFolded = !this.state.isFolded;
+    }
+}
+
+LayersPanel.template = "base_geoengine.LayersPanel";
```
ged-odoo, line 241: starting in odoo 16.1, we officially allow the use of `static` in component. I like that a lot, since it feels like the static props/template/components key belong to the top of the component code

base_geoengine/static/src/js/views/geoengine/layers_panel/layers_panel.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<templates>
+    <t t-name="base_geoengine.LayersPanel" owl="1">
+        <div
+            class="menu border-end"
+            t-attf-class="{{ state.isFolded ? 'o_layer_panel_fold ' : 'o_layer_panel bg-view' }}"
+        >
```
ged-odoo, line 7: not sure I like the formatting, but it seems quite readable indeed. is it the output of some tools like prettier?

base_geoengine/static/src/js/views/geoengine/records_panel/records_panel.esm.js
```diff
@@ -0,0 +1,77 @@
...
+        super.setup();
+
+        this.state = useState({
+            isFolded: false,
+            isClicked: 0,
+            modelDescription: "",
+            records: [],
+        });
+        this.orm = useService("orm");
+        onWillStart(() => (this.state.records = this.props.list.records));
```
ged-odoo, line 26: this feels like a red flag to me. usually, duplicating props value in component state is wrong. I however don't know enough of your code to be sure yet

base_geoengine/static/src/js/views/geoengine/records_panel/records_panel.esm.js
```diff
@@ -0,0 +1,77 @@
...
+        onWillUpdateProps((nextProps) => (this.state.records = nextProps.list.records));
+        onWillRender(async () => {
+            // Retrieves the name of the current model
+            const result = await this.orm.call("ir.model", "display_name_for", [
```
ged-odoo, line 30: be aware that the `onWillRender` hook is not asynchronous, so you will have a small flickering issue: owl will render immediately, then when the rpc comes back, it will render again

lmignon:
@kouffsamuel Unfortunately, the last commit create a circular dependency (sorry). You should remove-it to make fix CI.

base_geoengine/static/src/js/views/geoengine/records_panel/records_panel.esm.js
ged-odoo: maybe consider having `activeRecords`  getter in your class that simply return the props.records filtered by active filter, so you can remove `records` from your state (and probably need to add the filterValue to the state)

base_geoengine/static/src/css/style.css
```diff
@@ -0,0 +1,402 @@
...
+}
+
+.ol-scale-bar {
+    position: absolute;
+    bottom: 8px;
+    left: 8px;
+}
+
+.ol-scale-bar-inner {
+    display: flex;
```
ged-odoo, line 53: the guideline in odoo is now to prefer using boostrap css class above custom css. It helped a lot to reduce repetitive common css properties, such as display flex. So, for example, here, you could simply set `d-flex` on the element. (note that it may not work here, if you don't own the code that render this css class). Just saying that if you want, you can probably replace some of your css with bootstrap.

base_geoengine/static/src/js/widgets/geoengine_edit_map/field_geoengine_edit_map.esm.js
```diff
@@ -0,0 +1,327 @@
...
+
+export class FieldGeoEngineEditMap extends Component {
+    setup() {
+        super.setup();
+
+        // Allows you to have a unique id if you put the same field in the view several times
+        this.id = `map_${Date.now()}`;
+        this.orm = useService("orm");
+
+        onWillStart(() => Promise.all([this.loadJsFiles()]));
```
ged-odoo: you can now create a bundle and just load the bundle here

base_geoengine/static/src/js/widgets/domain_selector_geo_field/domain_field.esm.js
```diff
@@ -0,0 +1,34 @@
...
+        } catch (_e) {
+            // WOWL TODO: rethrow error when not the expected type
+            Object.assign(this.state, {recordCount: 0, isValid: false});
+            return;
+        }
+        Object.assign(this.state, {recordCount, isValid: true});
+    }
+}
+
+registry.category("fields").add("domain", DomainFieldExtend);
```
ged-odoo: you can use the force: true option, and not have to remove it from the registry before

base_geoengine/static/src/js/widgets/domain_selector_geo_field/domain_field.esm.js
```diff
@@ -0,0 +1,34 @@
+/** @odoo-module **/
+
+import {DomainField} from "@web/views/fields/domain/domain_field";
```
ged-odoo, line 3: be aware that we changed the domain selector a lot in master, so sadly, there will be some migration work to migrate this addon to 17.0 in the future

base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
```diff
@@ -0,0 +1,1273 @@
...
+        });
+
+        onPatched(() => {
+            if (this.map !== undefined && !this.state.isModified) {
+                this.renderVectorLayers();
+            }
+        });
+    }
+
+    async loadJsFiles() {
```
ged-odoo: you can just put all your js/css files in a new bundle, and load that bundle at once

ged-odoo (commented):
as I was pinged, I just did a quick review. I hope that's ok. This is only my comments after a quick look. This is way too big for me to have a serious look in 30min.

However, my overall impression is that it feels like a pretty good work. It involves so many different parts of the JS framework, and it looks like they are all used properly. So congratulation to everyone who worked on this.

lmignon:
@ged-odoo I'm so grateful for your review. Thank you!!!!

base_geoengine/static/src/js/views/geoengine/layers_panel/layers_panel.esm.js
```diff
@@ -30,15 +30,7 @@ export class LayersPanel extends Component {
...
-            this.isGeoengineAdmin = await this.user.hasGroup(
-                "base_geoengine.group_geoengine_admin"
-            );
-            const result = await this.orm.call(
-                this.props.model,
-                "get_geoengine_layers",
-                []
-            );
-            this.state.geoengineLayers = result;
+            await Promise.all([this.load()]);
```
lmignon: @kouffsamuel IMO your `Promise.all` is not well written. To allows // calls of the `user.hasGroup` and `orm.call` methods this should be written as (not sure but pretty):

```javascript

    await Promise.all([this.loadIsAdmin(), this.loadLayers() ]);
  });
}

async loadIsAdmin() {
  return this.user.hasGroup(
      "base_geoengine.group_geoengine_admin"
  ).then(function(result){
      this.isGeoengineAdmin = result; 
  });
}

async loadLayers () {
  return this.orm.call(
      this.props.model,
      "get_geoengine_layers",
      []
  ).then(function(result){
    this.state.geoengineLayers = result;
});
}

bealdav:
Hi thanks for your contributions, I don't understand here

#338 depends on #334

#334 depends on #338

How to solve it ?

lmignon:
ocabot merge nobump

PR #351, opened 2023-11-20
[12.0] [IMP] geoengine_swisstopo : Add Swisstopo projections

sbrunner:


cyrilmanuel (approved):
for the code, LGTM :) 

 but for the 3 commits, i'm not sure if you shall split them in 2 (pre-commit and swisstopo projections) 

same here for the name of the PR. i'm not sure if shall be rename to contain the version, type of pr (implementation) and a title explicit about the module that contain most of the modification and a description of the implementation. example  : 
[12.0] [IMP] base_geoengine : Adding Swisstopo projections 

@yvaucher could you confirm or refute this :))

geoengine_swisstopo/static/src/js/geoengine_swisstopo.js
```diff
@@ -1,124 +1,225 @@
...
+            init_proj4(this);
+        },
+        _render: function (parent) {
+            define_projections();
+            this._super.apply(this, arguments);
+        },
+
+    });
+});
+odoo.define("geoengine_swisstopo.projection_EPSG_2056", function (require) {
```
yvaucher: Seems unecessary to me to call `init_proj4` per projection.

geoengine_swisstopo/static/src/js/geoengine_swisstopo.js
```diff
@@ -1,124 +1,225 @@
...
+            this._super.apply(this, arguments);
+            init_proj4(this);
+        },
+        _render: function (parent) {
+            define_projections();
+            this._super.apply(this, arguments);
+        },
+    });
+});
+odoo.define("geoengine_swisstopo.projection_EPSG_21781", function (require) {
```
yvaucher: To rename `geoengine_swisstopo.init_projections` ? And have a single call to define_projections and init_proj4.

geoengine_swisstopo/static/src/js/geoengine_swisstopo.js
```diff
@@ -1,124 +1,225 @@
...
 
-/**
- * Extents of Swiss projections. (EPSG:21781)
- */
-var EXTENT = [420000, 30000, 900000, 350000];
+odoo.define("geoengine_swisstopo.projection_EPSG_4326", function (require) {
+    "use strict";
 
-var PROJECTION_CODE = "EPSG:21781";
+    const GeoengineWidgets = require("base_geoengine.geoengine_widgets");
```
yvaucher, line 72: The `require` is similiar as an `import` wouldn't it make more sense to keep it at the root level of the file or was there some cyclic import issue?

geoengine_swisstopo/geo_view/geo_raster_layer.py
```diff
@@ -1,7 +1,15 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+import logging
+
+import requests
+from odoo import api, fields, models
+
+_LOGGER = logging.getLogger(__name__)
+_CAPABILITIES_URL = (
```
yvaucher, line 11: As those 2 variables are defined in the last commit, it seems the splitting of commit is odd.

yvaucher (changes requested):
Looks great, just have a little concern on the multiple odoo.define for projections as it seems redundant.

Plus about the commits, some part of the changes of the 2nd commit deserve to be in the 3rd commit (or be squashed) the linting could go in a separate commit to let your last commit emphasis on the practical changes.

geoengine_swisstopo/geo_view/geo_raster_layer.py
```diff
@@ -5,8 +5,53 @@
...
+                    _LOGGER.error(
+                        "Swisstopo WMTS Capabilities request (%s)\n"
+                        "failed with status code %s:\n%s",
+                        url,
+                        response.status_code,
+                        response.text,
+                    )
+                    record.capabilities = False
+            else:
+                record.capabilities = False
```
yvaucher: This doesn't looks like it should be part of the pre-commit

sbrunner: Done, tanks :-)

geoengine_swisstopo/static/src/js/geoengine_swisstopo.js
sbrunner: I just keep it as it was, and the require function is not defined at top level...

geoengine_swisstopo/geo_view/geo_raster_layer.py
sbrunner: Looks to be fixed in the last version :-)

yvaucher (approved):

cyrilmanuel (approved):

base_geoengine/geo_model.py
```diff
@@ -131,7 +131,7 @@ def get_edit_info_for_geo_column(self, column):
             'default_extent': view.default_extent or DEFAULT_EXTENT,
             'default_zoom': view.default_zoom,
         }
-        logger.debug("Parameters for geo field {}:\n{}".format(column, res))
+        logger.debug(f"Parameters for geo field {column}:\n{res}")
```
HekkiMelody, line 134: the minimum version of python supported by v12 is 3.5, which does not have fstrings. This unfortunately broke my installation.

.pre-commit-config.yaml
```diff
@@ -56,4 +56,45 @@ repos:
...
+          - --py36-plus
+  # - repo: https://github.com/PyCQA/isort
+  #   rev: 5.12.0
+  #   language_version: python3.8
+  #   hooks:
+  #     - id: isort
+  #       name: isort except __init__.py
+  #       args:
+  #         - --settings=.
+  #       exclude: /__init__\.py$
```
HekkiMelody, line 100: This shouldn't be modified manually, but only through `copier update`

HekkiMelody:
Fixed the issue in https://github.com/OCA/geospatial/pull/366

PR #342, opened 2023-06-29
[16.0][mig] web_view_leaflet_map and web_view_leaflet_map_partner

legalsylvain:
- Trivial port of the modules : ``web_view_leaflet_map`` and ``web_view_leaflet_map_partner``

- mig commit to review : https://github.com/OCA/geospatial/pull/342/commits/1e873b609579f85f8520dfd5da6668c8353931b6

- bump leaflet lib from 1.8.0 , released on April 18, 2022 to 1.9.4, released on May 18, 2023.

res_partner_map

PR #352, opened 2023-11-30
[16.0][MIG] web_view_leaflet_map

yibudak:
Since web_map module is only exist for enterprise edition, I wanted to port OpenStreetMap viewer module to Odoo 16.0. Also I've made some changes to improve module quality. Here are the changes:

Migrated the module.
Updated the Leaflet library to the latest version.
Added a default marker.
Corrected the current partner location issue (previously taking the active company's location).
Fixed compatibility issue with the web_responsive module.
Refactored some code related to popups to improve readability.
Resolved the error where the map does not re-render when exiting the form view.


yibudak:
@pedrobaeza @legalsylvain @yvaucher

pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

And run pre-commit locally for generating the changes, instead of manually do changes in README.md, and the link for the setup folder (which is now incorrect).

About the module itself, I can't say, as I don't use it.

legalsylvain:
hi @yibudak. 
First, thanks for porting this module ! 
Same remarks as @pedrobaeza for the history.
could you also port this trivial module in the current PR ? https://github.com/OCA/geospatial/tree/12.0/web_view_leaflet_map_partner
It allow to test easily on runbaot the feature. 

Please ping me when it's ready for review.
kind regards.

Also 2 questions : 

>  Corrected the current partner location issue (previously taking the active company's location).

Could you explain what is now how is defined the default location ? 

>  Resolved the error where the map does not re-render when exiting the form view.

Thanks ! 

Note : I began a work that is to review here : https://github.com/OCA/geospatial/pull/342

could you base your changes on my unreviewed PR ?

yibudak:
Hey @legalsylvain, 

I didn't see your work about migrating these modules into v16. What do you mean by "could you base your changes on my unreviewed PR" ? Should I close this PR and commiting into your migration branch?

Corrected the current partner location issue (previously taking the active company's location).

It was like this: 
` 
       current_partner = self.env.user.company_id.partner_id
`
And I've changed it into:
`
        current_partner = self.env.user.partner_id
`        
This is because company employees (users) can be located at different places. if it was like before, only the location defined to the company would appear.

And the second question. Actually that "invalidateSize with timeout" is a correct and quick workaround that problem, I just increased the timeout because it was too low (1 ms) and it wasn't working as expected.

legalsylvain:
> Should I close this PR and commiting into your migration branch?

Maybe we can review and merge #342 as it is working as in V12, then you can make a little IMP/FIX PR against 16.0 branch. What do you think ?

yibudak:
> > Should I close this PR and commiting into your migration branch?
> 
> Maybe we can review and merge #342 as it is working as in V12, then you can make a little IMP/FIX PR against 16.0 branch. What do you think ?

Agreed. I will make contributions into your work. Thank you for your quick response 👍🏻

PR #355, opened 2024-01-19
Fix map render

igr-ok:
Fix map render - change timeout value.


legalsylvain (commented):
Hi @igr-ok. Thanks for your PR !

Question : This look quite high, and can generate delay for end user. Did you tried something smaller, like 100 ? 

Technical : could you squash commits ? 

regards.

igr-ok:
Hi @legalsylvain, yes  500ms too much, 100 is OK i cheked;

legalsylvain:
Hi @igr-ok. Thanks could you change the value and squash commits ? thanks !

igr-ok:
@legalsylvain, ok

PR #359, opened 2024-02-22
[ADD] MIG geoengine_base_geolocalize to v9

gfcapalbo:
Tested and verified, ported to V9

### DIFF
--- a/geoengine_base_geolocalize/__openerp__.py
+++ b/geoengine_base_geolocalize/__openerp__.py
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 {'name': 'Geospatial support for base_geolocalize',
- 'version': '8.0.0.1.0',
+ 'version': '9.0.0.1.0',
  'category': 'GeoBI',
  'author': "ACSONE SA/NV, Odoo Community Association (OCA)",
  'license': 'AGPL-3',
@@ -35,8 +35,8 @@
  'data': [
      'views/res_partner_view.xml'
  ],
- 'installable': False,
+ 'installable': True,
  'application': True,
- 'autoinstall': True,
- 'active': False,
+ 'autoinstall': False,
+ 'active': True,
  }
--- a/geoengine_base_geolocalize/models/res_partner.py
+++ b/geoengine_base_geolocalize/models/res_partner.py
@@ -71,6 +71,8 @@ class ResPartner(geo_model.GeoModel):
 
     @api.one
     def geo_localize(self):
+        # overwriting completely the standard geo_localize in base_geolocation
+        # wich uses google API key. this uses openstreetmap.
         self.geocode_address()
         return True
 


NL66278 (changes requested):
Migrations should be with history.

gfcapalbo:
> Migrations should be with history.
@NL66278 i read the migration guide v8-v9: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0  and unfortunately i don't understand. 
This module was already in v9.0 branch as-is- and i just checked all parts , read it , made it active, and tested it.
It has an identical history of v9.0  + 1 commit.  The history is fully there.  It is not taken from v8.0 branch.

The history and commits that involved this modulare already in the v9.0 branch ,
If i extracted all the commits from V8 where the module was changed, I would be adding commits that are redundant with :
https://github.com/OCA/geospatial/commits/9.0/geoengine_base_geolocalize

You can see here the full history of this module, wich is already on branch v9.0

Just go on  the module folder , and use "history" feature, you will see 14 commits already there in v9.0 branch involving this module:


Screenshot_2024-02-23_10-10-14

NL66278 (approved):
:+1: LGTM I was confused by seeing only one commit.

gfcapalbo:
> 👍 LGTM I was confused by seeing only one commit.

My request for review was more about not understanding why this module was deactivated, the commits give no reason. it is 1 small peice of python code, and 1 small view change, with tests.
I reviewed it too and it looks fine.  On test server it worked fine geolocating correctly 2200 partners. @pedrobaeza  was this left deactivated only by default, and then never tested and approved by anybody?

pedrobaeza:
It seems in this repository, they took the approach of just setting the modules as installable=False instead of erasing whole commit history.

/ocabot merge major

yvaucher:
Let's try this again, I don't see any status on previous attempt to merge

/ocabot merge major

yvaucher (approved):

yvaucher:
CI is not triggered by ocabot on the old 9.0 branch?

Let's do it manually then.

gfcapalbo:
@yvaucher   @pedrobaeza @NL66278   Please revert this MR.  I have  some troubling findings.


The Base_geoengine  geo_edit_map  widget does not work with the parameters returned by this geolocation.


To Reproduce:
        1 .Geolocate the customer with  geoengine_base_geolocalize  Long/lat will be written to partner.
        2. insert long/lat in nominatim website and you will see the correct point.
        3. when displaing this point via widget = "geo_edit_map", we will have a display of a point , usually several KM away in both latitude and longitude. 
      
It is clear that the current presentation layer is not compatible with the data returned by this geolocation. We have a customer  database with old geolocation data that , with different LONG/LAT, displays the correct point for that customer.   I am not sure what it may be, we suspect it may be projection (https://gis.stackexchange.com/questions/48949/epsg-3857-or-4326-for-web-mapping)
 but i have not found a way to make the data returned by nominatim compatible with this older widget ( or viceversa)


  Currently as it stands all results from geolocation will be displayed wrong with current  widget="geo_edit_map", so i propose to revert this MR, in order not to cause problems to others. 

Will also create an issue for this. thank you

pedrobaeza:
Well, the module didn't exist yet in v9 before this PR, so I think it's better to simply put a PR with the fixes.

gfcapalbo:
> Well, the module didn't exist yet in v9 before this PR, so I think it's better to simply put a PR with the fixes.

@pedrobaeza I would post fixes, but i have none.  after some hours of reading and work, i cannot figure out why is the "geo_edit_map" displaying the results of  the geolocation incorrectly, while  https://nominatim.openstreetmap.org/ui/reverse.html  shows the perfect spot using the same coordinates.

I have excluded any rounding errors possibilities.


We might say that the "bug" is not here , but in  the geo_engine_map,  https://github.com/OCA/geospatial/blob/9.0/base_geoengine/static/src/js/views/geoengine_widgets.js
 but i cannot be certain for now.

therefore i propose to revert and leave the MR open, asking for help from a geolocation expert.

pedrobaeza:
I leave also the decision to the current PSCs. I think this doesn't hurt at all merged. We just can do the mention in the README. As PR, nobody is going to check it.

PR #361, opened 2024-02-23
[16.0][FIX] geospatial_partner - error message on no partner with valid geom

wouitmil:
Added two fixes to avoid error message when no partners are encoded with a valid geo_point.

1. The check extent !== infinite_extent wasn't working as it's not a recommended way to compare arrays in javascript.
2. the filter in the right panel wasn't working if no partners are in the list

Added the geoengine view in the default contacts/resp_partner action


cyrilmanuel (approved):
LGTM until pre-commit fix

wouitmil:
> LGTM until pre-commit fix

I'll try to fix it. It may be caused by a working syntax (optional chaining) that is not accepted by this version of eslint. Is it possible that the version is too old?

I see [here](https://github.com/eslint/eslint/issues/13566) that it's should be fixed since eslint v7.9.0 (and I see that the version used is v8.24.0)

wouitmil:
> LGTM until pre-commit fix

Fixed ;-)

yvaucher (approved):

yvaucher:
Can you improve the commit message `[FIX] base_geoengine & geospatial_partner `

Question to answer is: What do you fix in those modules?

PR #386, opened 2024-11-03
[18.0][MIG] base_geoengine

weinni2000:
Migration to 18.0

I tried to make the migration but I have troubles with the GeoengineRecord. I'm not very firm with owl so maybe some of my questions are not that precise.
But somehow it doesn't pick up the template for the GeoengineRecord
If I add it within the js it works. 
I still get this error:
Uncaught (in promise) Error: Missing template: "web.Field" (for component "Field")
I'm a little bit lost because I don't know how to find the templates referenced by eg. t-call ... __template__25.
Would be great if someone could help me with that.


yvaucher:
Thanks for the attempt of migration.

Can you provide a stack trace of the error?

`web.Field`  template should be present as it's comming from the core addon `web`:
https://github.com/odoo/odoo/blob/18.0/addons/web/static/src/views/fields/field.xml

I would assume that `__template__25` is an automatically generated template. Could it be what you have for instance in 
https://github.com/OCA/geospatial/pull/386/commits/e90815b25ed6f3f3545980eb05e9cbf6fc492a5e#diff-edc81af2df248b5194f57bb65b9fb3fb5b5488961d022dd98515218af97f0157R32-R43 ?

Otherwise I notice the CI fails as you are missing requirements like shapely, you will want to ensure you have the same deps as in: 
https://github.com/OCA/geospatial/blob/17.0/requirements.txt

(IIRC pre-commit should do it for you)

weinni2000:
@yvaucher:
I made a new attempt:
How does it work to add the apt dependencies in the CI Pipline? 
In my doodba environment I added the following apt:txt:
postgis
gdal-bin
libgdal-dev
g++

ap-wtioit:
> @yvaucher: I made a new attempt: How does it work to add the apt dependencies in the CI Pipline? In my doodba environment I added the following apt:txt: postgis gdal-bin libgdal-dev g++

those are dependencies for the postgres container used in the test not for the OCA module your are migrating. It seems you need to change the postgres services used by the github pipeline as it was done in the 17.0 migration: https://github.com/OCA/geospatial/commit/51426372e4a8638493b009be7ae23078892450ad#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88L45

weinni2000:
I still have an issue. If I open a Popup:

Bildschirmfoto 2024-11-13 um 11 51 35

The templates can't be loaded:
If I change the named templates
_template: "module.name"_
to an inline xml 
_template: xml`xxxxxx`_

it works. This happens in the geoengine_renderer.js and downstream (like in web.Field).

weinni2000:
@yvaucher:
The tests are working now.
To properly extend odoo I had to override the parse method in odoo.osv.expressions. If my approach makes sense I would create an PR at odoo to give us an entrypoint so I don't have to override to whole method (And to make them aware of other uses for this parsers). Could somebody have a look at this first because this for sure will take some time.

LuvForAirplanes:
@weinni2000 bumping this one again. Did Odoo get back to you about getting an additional entry point here?

weinni2000:
@LuvForAirplanes:
It works for my purpose now but as the maintainers didn't really care to have a look I thought it's best to wait till it get's interesting for somebody else too.

yterrettaz (approved):
I performed a simple test on an object with a form view and a list view. Raster and layer configuration. Everything works fine for me

yvaucher:
While this PR was the oldest I'm closing it in favor of #403 where the expressions are better integrated.

Thanks for sharing your work, nevertheless.

PR #393, opened 2025-01-31
[16.0][FIX] geoengine_base_geolocalize: Extend the correct view to avoid problems when installing the module

CarlosRoca13:
Sometimes, when installing the module, an error related to its installation occurs. To prevent this random error, the view that adds the page is extended in such a way that there is no possibility for the error to occur.

cc @Tecnativa 

ping @sergio-teruel @carlosdauden


carlosdauden (approved):

amh-mw (approved):
Code review only. LGTM.

PR #400, opened 2025-04-17
[16.0][IMP][base_geoengine]: Update Domain for attribute_field_id

anusriNPS:
Updated domain such that attribute_field_id display values related to "geo_field_id.model_id"

### DIFF
--- a/base_geoengine/models/geo_vector_layer.py
+++ b/base_geoengine/models/geo_vector_layer.py
@@ -61,9 +61,12 @@ class GeoVectorLayer(models.Model):
         ondelete="cascade",
         domain=[("ttype", "ilike", "geo_")],
     )
-    attribute_field_id = fields.Many2one(
-        "ir.model.fields", "Attribute field", domain=[("ttype", "in", SUPPORTED_ATT)]
+
+    attribute_field_id_domain = fields.Binary(
+        compute="_compute_attribute_field_id_domain", readonly=True, store=False
     )
+    attribute_field_id = fields.Many2one("ir.model.fields", "Attribute field")
+
     model_id = fields.Many2one(
         "ir.model",
         "Model to use",
@@ -154,3 +157,15 @@ class GeoVectorLayer(models.Model):
                     rec.model_id = ""
             else:
                 rec.model_id = ""
+
+    @api.depends("geo_field_id")
+    def _compute_attribute_field_id_domain(self):
+        for rec in self:
+            rec.attribute_field_id_domain = (
+                [
+                    ("ttype", "in", SUPPORTED_ATT),
+                    ("model", "=", rec.geo_field_id.model_id.model),
+                ]
+                if rec.geo_field_id
+                else [("ttype", "in", SUPPORTED_ATT)]
+            )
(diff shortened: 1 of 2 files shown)


base_geoengine/models/geo_vector_layer.py
```diff
@@ -154,3 +157,25 @@ def _compute_model_id(self):
...
+                [
+                    ("ttype", "in", SUPPORTED_ATT),
+                    ("model", "=", rec.geo_field_id.model_id.model),
+                ]
+                if rec.geo_field_id
+                else [("ttype", "in", SUPPORTED_ATT)]
+            )
+
+    @api.onchange("geo_field_id")
+    def _onchange_geo_field_id(self):
```
SirPyTech: Returning domains from an `onchange` is deprecated, see https://github.com/odoo/odoo/blob/10ce67f5cad59d0a6b4877cd377d4a8341480052/odoo/models.py#L6365.
We should be able to do everything only with the `domain` attribute, see the examples in the PR https://github.com/odoo/odoo/pull/41918 that added the deprecation.

base_geoengine/models/geo_vector_layer.py
```diff
@@ -154,3 +157,25 @@ def _compute_model_id(self):
...
+        return {
+            "domain": {
+                "attribute_field_id": [
+                    ("ttype", "in", SUPPORTED_ATT),
+                    ("model", "=", self.geo_field_id.model_id.model),
```
SirPyTech: Looking at the _computed_ code, I think this term should be added only when `self.geo_field_id` is _truthy_.

SirPyTech (commented):
Thanks for the PR!

base_geoengine/models/geo_vector_layer.py
SirPyTech: Maybe the issue you are facing is that the *computed* domain does not recompute when needed: in this case you can try adding `api.depends` to it to trigger its recomputation.

anusriNPS: Using api.depends based on "geo_field_id" for binary field able to display domains related to geo_field_id

anusriNPS:
When geo_field_id is using "res.partner" for coordinates, attribute_field_id values are displayed as below
Screenshot from 2025-04-23 08-38-31


When using "res.users" for coordinates,
Screenshot from 2025-04-23 08-38-22

base_geoengine/models/geo_vector_layer.py
SirPyTech: :+1:

base_geoengine/models/geo_vector_layer.py
SirPyTech: :+1:

SirPyTech (approved):
Thanks

HekkiMelody (approved):
Code review, LGTM

SirPyTech:
@OCA/geospatial-maintainers maybe someone wants to have a look? We already have 2 approvals, thanks!

yvaucher:
Thanks for the changes

yvaucher (approved):

PR #403, opened 2025-06-04
[18.0][MIG] base geoengine: Migration to 18.0

cormaza:
Standard migration, refactored some methods with new version

Supersed: https://github.com/OCA/geospatial/pull/386


yterrettaz (approved):
I am testing this pull request from a functional perspective within the Odoo GeoEngine module. It includes geographic views on a specific object using OSM raster layers. Projections, spatial filters, and geometry types (polygon and point) have been successfully verified during my tests

Scuottolinx:
Hi, I tested the plugin, and it doesn't behave as it should.
If I create combined vectors using the model_id and model_domain attributes, the result isn't 100% correct:
- the points on the map are displayed correctly (OK)
- the detail of individual points, when clicking on a point, is always the same, regardless of the selected point (KO)

Any idea why?

Furthermore, the vector filter in the interface displays this error message when clicked:
`Uncaught Promise > Invalid props for component 'DomainSelector': 'domain' is undefined (should be a string)
`

Scuottolinx:
Hi, I solved my problem:

> The detail of individual points, when clicking on a point, is always the same, regardless of the selected point (KO).

 just add the <field name="id"/> field to <geoengine>

> Uncaught Promise > Invalid props for component 'DomainSelector': 'domain' is undefined (should be a string).

This error is due to the fact that the Dialog in Odoo 18 has changed, and therefore in:

- layers_panel.esm.js, I had to add the "domain" and "onConfirm" attributes.
- In the DomainSelectorGeoFieldDialog, I obviously defined the same properties.

Maybe it can be useful for a new commit :)

cormaza:
> Hi, I solved my problem:
> 
> > The detail of individual points, when clicking on a point, is always the same, regardless of the selected point (KO).
> 
> just add the field to
> 
> > Uncaught Promise > Invalid props for component 'DomainSelector': 'domain' is undefined (should be a string).
> 
> This error is due to the fact that the Dialog in Odoo 18 has changed, and therefore in:
> 
> * layers_panel.esm.js, I had to add the "domain" and "onConfirm" attributes.
> * In the DomainSelectorGeoFieldDialog, I obviously defined the same properties.
> 
> Maybe it can be useful for a new commit :)

I will review and back with solution

yvaucher (approved):
LGTM code review only

Great work there, and indeed seems more integrated than #386  for the expressions adaptations to odoo core.

PR #406, opened 2025-07-27
18.0 mig base geoengine demo

weinni2000:
Standard Migration:

If I'm not in Developer Mode I get:
Caused by: ReferenceError: ol is not defined

Bildschirmfoto 2025-07-27 um 12 41 25

In Devoloper Mode everything works


cparadis-impressfoods (changes requested):
LGTM except the history could be squashed and cleaned up.

PR #409, opened 2025-08-07
[16.0][IMP] base_geoengine: Include edit button in Popup

anusriNPS:
Added edit button in popup of record shown in geoengine view which allows user to edit records by opening new tab in browser window.

### DIFF
--- a/base_geoengine/static/src/js/views/geoengine/geoengine_controller/geoengine_controller.esm.js
+++ b/base_geoengine/static/src/js/views/geoengine/geoengine_controller/geoengine_controller.esm.js
@@ -65,6 +65,26 @@ export class GeoengineController extends Component {
     }
 
     /**
+     * Allow you to edit form view for the filled-in model
+     * from new tab in the browser window.
+     * @param {*} resModel
+     * @param {*} resId
+     */
+    async editRecord(resModel, resId) {
+        const url = window.location.href;
+        const searchParams = new URLSearchParams(url.split("#")[1]);
+        searchParams.set("view_type", "form");
+        searchParams.set("id", resId);
+        searchParams.set("menu_id", searchParams.get("menu_id"));
+        const action_url = url.split("#")[0] + "#" + searchParams.toString();
+        this.actionService.doAction({
+            type: "ir.actions.act_url",
+            target: "_blank",
+            url: action_url,
+        });
+    }
+
+    /**
      * When you finished drawing a new shape, this method is called to open form view and create the record.
      * @param {*} resModel
      * @param {*} field
--- a/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.xml
+++ b/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.xml
@@ -14,6 +14,11 @@
             />
         </div>
         <div id="popup" class="ol-popup">
+             <div
+                id="popup-editor"
+                class="ol-popup-editor text-primary"
+                t-on-click="onEditButtonClicked"
+            />
             <div
                 id="popup-closer"
                 class="ol-popup-closer text-primary"
--- a/base_geoengine/static/src/js/views/geoengine/geoengine_controller/geoengine_controller.xml
+++ b/base_geoengine/static/src/js/views/geoengine/geoengine_controller/geoengine_controller.xml
@@ -29,6 +29,7 @@
                 data="model.root"
                 editable="editable"
                 openRecord.bind="openRecord"
+                editRecord.bind="editRecord"
                 updateRecord.bind="updateRecord"
                 onClickDiscard.bind="onClickDiscard"
                 createRecord.bind="createRecord"
--- a/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
+++ b/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
@@ -632,6 +632,14 @@ export class GeoengineRenderer extends Component {
     }
 
     /**
+     * When you click on the edit button, it calls the controller's
+     * editRecord method.
+     */
+    onEditButtonClicked() {
+        this.props.editRecord(this.record.resModel, this.record.resId);
+    }
+
+    /**
      * Allows you to change the visibility of layers. This method is called
      * when the user changes raster layers.
      */
@@ -1256,6 +1264,7 @@ GeoengineRenderer.props = {
     archInfo: {type: Object},
     data: {type: Object},
     openRecord: {type: Function},
+    editRecord: {type: Function},
     editable: {type: Boolean, optional: true},
     updateRecord: {type: Function},
     onClickDiscard: {type: Function},
--- a/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.scss
+++ b/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.scss
@@ -65,3 +65,14 @@
 .ol-popup-closer:after {
     content: "✖";
 }
+.ol-popup-editor {
+    text-decoration: none;
+    position: absolute;
+    top: 1px;
+    right: 25px;
+}
+.ol-popup-editor:after {
+    font-family: FontAwesome;
+    content: "\f040";
+    cursor: pointer;
+}
--- a/geoengine_base_geolocalize/tests/test_geoengine_partner.py
+++ b/geoengine_base_geolocalize/tests/test_geoengine_partner.py
@@ -1,5 +1,7 @@
 # Copyright 2015-2017 ACSONE SA/NV (<http://acsone.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from unittest.mock import MagicMock, patch
+
 from odoo.tests.common import TransactionCase
 
 
@@ -18,7 +20,14 @@ class TestGeoenginePartner(TransactionCase):
         partner_id.partner_longitude = 20
         self.assertTrue(partner_id.geo_point, "Should have geo_point")
 
-    def test_geo_localize(self):
+    @patch("requests.get")
+    def test_geo_localize(self, requests_get):
+        mock_response = MagicMock()
+        mock_response.status_code = 200
+        mock_response.json = MagicMock(
+            return_value=[{"lat": "49.9549071", "lon": "5.4085830"}]
+        )
+        requests_get.return_value = mock_response
         vals = {
             "name": "Partner Project",
             "street": "Rue au bois la dame",


cparadis-impressfoods (approved):
LGTM!

base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.xml
```diff
@@ -14,6 +14,12 @@
             />
         </div>
         <div id="popup" class="ol-popup">
+             <div
+                t-if="isGeoengineAdmin"
+                id="popup-editor"
+                class="ol-popup-editor text-primary"
+                t-on-click="onEditButtonClicked"
+            />
```
quirino95, line 21: I have some observations about this:
- Is really GeoenginAdmin check needed? I think is more correct that the check should be done on the work order itself (user could be GeoengineAdmin but not allowed to edit a specific WO, how is this situation managed?);
- When moving the cursor over the button, it doesn't give me the impression that it's clickable. Please could you improve this by adding a `clickable` attribute or changing the `div` into a `button` or something similar?
- When clicking on `edit` button, the form view just opened is "full screen", hiding the upper navigation bar, and when going back, map view is reset;
- **suggestion**: clicking the "open" button, i see the same form that I have by clicking on "edit" button. What if the open button opens form view in edit mode (depending of access rules of course)? In that case, "edit" button would become obsolete, but the interface should become more clear;

HekkiMelody: Re: the first point, this is a generic geoengine module, it could be used with any object, not just fieldservice. But in general, I agree with checking permissions on the object itself if possible and not requiring admin rights on the Geoengine app.

If it's too complicated, Odoo probably already checks permissions in other ways so I'd rather leave both buttons for everyone, at least as a first implementation.

anusriNPS: 1. Regarding edit button, icon is shown as clickable attribute when hovered around it. Edit button opens url in new tab with navigation bar and also restores history in geoengine view.
2. Regarding permission on object level:
     a. Tried to hide the editbutton using ormservice and reading permission for record from backend during onwillstart() in geoengineRenderer. However, each record related information is attached in geoengineRenderer only when displayPopupRecord() method is called and geoenginerecord is mounted based on information passed from recordPanel template. Component is getting destroyed while trying to fetch specific record from geoengineRenderer setup. So, could not hide the button completely. This would be best option for this case but could not achieved.
     b. Currently, if a model has a field named "can_write", if passed in info_box template from geoengine view definition, record is validated based on can_write field and allows user to edit the record. If not, user is notified that sufficient permissions are not available to edit the record. If info_box template does not have "can_write" field, all users is allowed to edit the record.(default behaviour).
     
Note:  Without approach "b", by default even when user is allowed to edit the record, permissions are validated in backend and user is blocked in case the user does not have sufficient permissions.

IMO, I would like to stick not using approach "b" as already backend permissions are validated even without exposing new field in model for this purpose which seems to be overhead. 

Please let me know your views.

quirino95: Changes on the edit button now are ok.
About permissions, since now the form opens in a new tab, is ok for me to remove new checks added in this PR and stay with the "default" ones when opening the new tab.

anusriNPS: Removed checking permission based on "can_write" field introduced from model and preserved default behaviour for permission check of geoengine views for edit button.

quirino95 (approved):

base_geoengine/static/src/js/views/geoengine/geoengine_controller/geoengine_controller.esm.js
```diff
@@ -64,6 +64,31 @@ export class GeoengineController extends Component {
...
+            String(resId) +
+            "&view_type=form&model=" +
+            resModel +
+            "&menu_id=" +
+            searchParams.get("menu_id");
+        this.actionService.doAction({
+            type: "ir.actions.act_url",
+            target: "_blank",
+            url: action_url,
+        });
```
HekkiMelody: **chore**: not at all a fan of building the URL manually, but if there's no other easy way to do it, that's fine.

I would suggest building the `action_url` in a less "manual" way, like in this module https://github.com/OCA/web/blob/16.0/web_widget_open_tab/static/src/js/open_tab_widget.esm.js#L12-L25

Also, please use `const` and `let` instead of `var`.

HekkiMelody (commented):
Small code cleanup suggestion, otherwise seems ok.

base_geoengine/static/src/js/views/geoengine/geoengine_controller/geoengine_controller.esm.js
anusriNPS: Updated as suggested

HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
@OCA/geospatial-maintainers We've been using this successfully for a few months now. I would be grateful if you could review this. Thanks

legalsylvain:
No review, but as it is validated by 3 people. merging.

/ocabot merge minor

thanks for your contribution.

HekkiMelody:
I see that the error is due to

```
2026-01-26 11:26:30,614 334 INFO odoo odoo.addons.base_geolocalize.models.base_geocoder: openstreetmap nominatim service called 
2026-01-26 11:26:30,615 334 WARNING odoo odoo.addons.base_geolocalize.models.base_geocoder: Request to openstreetmap failed.
Code: 403
Content: b'<html>\n<head>\n<title>Access blocked</title>\n</head>\n<body>\n<h1>Access blocked</h1>\n\n<p>You have been blocked because you have violated the\n<a href="https://operations.osmfoundation.org/policies/nominatim/">usage policy</a>\nof OSM\'s Nominatim geocoding service. Please be aware that OSM\'s resources are\nlimited and shared between many users. The usage policy is there to ensure that\nthe service remains usable for everybody.</p>\n\n<p>Please review the terms and make sure that your\nsoftware adheres to the terms. You should in particular verify that you have set a\n<b>custom HTTP referrer or HTTP user agent</b> that identifies your application, and\nthat you are not overusing the service with massive bulk requests.\nIf you do that then the block usually goes away after some hours of inactivity.</p>\n\n<p>Please be aware that we are providing this service as-is and without any\nguarantees. We cannot provide support. If you need that, please have a look\nat comercial alternatives (also listed on the usage policy page) or set up\nyour own instance of Nominatim.</p>\n</body>\n</head>\n' 
2026-01-26 11:26:30,617 334 INFO odoo odoo.addons.geoengine_base_geolocalize.tests.test_geoengine_partner: ====================================================================== 
2026-01-26 11:26:30,617 334 ERROR odoo odoo.addons.geoengine_base_geolocalize.tests.test_geoengine_partner: ERROR: TestGeoenginePartner.test_geo_localize
Traceback (most recent call last):
  File "/opt/odoo/addons/base_geolocalize/models/base_geocoder.py", line 95, in _call_openstreetmap
    result = response.json()
  File "/opt/odoo-venv/lib/python3.10/site-packages/requests/models.py", line 900, in json
    return complexjson.loads(self.text, **kwargs)
  File "/opt/odoo-venv/lib/python3.10/site-packages/simplejson/__init__.py", line 514, in loads
    return _default_decoder.decode(s)
  File "/opt/odoo-venv/lib/python3.10/site-packages/simplejson/decoder.py", line 386, in decode
    obj, end = self.raw_decode(s)
  File "/opt/odoo-venv/lib/python3.10/site-packages/simplejson/decoder.py", line 416, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

```

we may be hitting openstreetmaps a bit too often? We may have to refactor the tests to not call OSM, or skip some tests in case of `Access Denied`

anusriNPS:
> I see that the error is due to
> 
> ```
> 2026-01-26 11:26:30,614 334 INFO odoo odoo.addons.base_geolocalize.models.base_geocoder: openstreetmap nominatim service called 
> 2026-01-26 11:26:30,615 334 WARNING odoo odoo.addons.base_geolocalize.models.base_geocoder: Request to openstreetmap failed.
> Code: 403
> Content: b'<html>\n<head>\n<title>Access blocked</title>\n</head>\n<body>\n<h1>Access blocked</h1>\n\n<p>You have been blocked because you have violated the\n<a href="https://operations.osmfoundation.org/policies/nominatim/">usage policy</a>\nof OSM\'s Nominatim geocoding service. Please be aware that OSM\'s resources are\nlimited and shared between many users. The usage policy is there to ensure that\nthe service remains usable for everybody.</p>\n\n<p>Please review the terms and make sure that your\nsoftware adheres to the terms. You should in particular verify that you have set a\n<b>custom HTTP referrer or HTTP user agent</b> that identifies your application, and\nthat you are not overusing the service with massive bulk requests.\nIf you do that then the block usually goes away after some hours of inactivity.</p>\n\n<p>Please be aware that we are providing this service as-is and without any\nguarantees. We cannot provide support. If you need that, please have a look\nat comercial alternatives (also listed on the usage policy page) or set up\nyour own instance of Nominatim.</p>\n</body>\n</head>\n' 
> 2026-01-26 11:26:30,617 334 INFO odoo odoo.addons.geoengine_base_geolocalize.tests.test_geoengine_partner: ====================================================================== 
> 2026-01-26 11:26:30,617 334 ERROR odoo odoo.addons.geoengine_base_geolocalize.tests.test_geoengine_partner: ERROR: TestGeoenginePartner.test_geo_localize
> Traceback (most recent call last):
>   File "/opt/odoo/addons/base_geolocalize/models/base_geocoder.py", line 95, in _call_openstreetmap
>     result = response.json()
>   File "/opt/odoo-venv/lib/python3.10/site-packages/requests/models.py", line 900, in json
>     return complexjson.loads(self.text, **kwargs)
>   File "/opt/odoo-venv/lib/python3.10/site-packages/simplejson/__init__.py", line 514, in loads
>     return _default_decoder.decode(s)
>   File "/opt/odoo-venv/lib/python3.10/site-packages/simplejson/decoder.py", line 386, in decode
>     obj, end = self.raw_decode(s)
>   File "/opt/odoo-venv/lib/python3.10/site-packages/simplejson/decoder.py", line 416, in raw_decode
>     return self.scan_once(s, idx=_w(s, idx).end())
> simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
> ```
> 
> we may be hitting openstreetmaps a bit too often? We may have to refactor the tests to not call OSM, or skip some tests in case of `Access Denied`

Will check on the issue

quirino95 (approved):
LGTM

Thank you for the fix!

HekkiMelody (approved):
Code review, LGTM for the test mock too.

Thank you!

HekkiMelody:
@OCA/geospatial-maintainers we've fixed the issue with the flaky tests (even though they were out of scope for this PR) could you try the merge again? Thank you

We can forward port both changes to the newest versions too

legalsylvain:
thanks for the contribution ! 

/ocabot merge minor

HekkiMelody:
Thank you @legalsylvain !

PR #411, opened 2025-08-29
[17.0][REF] base_geoengine: streamline imports and improve onWillStart logic in RecordsPanel

zamberjo:
I've also removed a rather annoying console.log.


zamberjo:
What do you think about this change? @kouffsamuel @peluko00 This way, a call is not made every time it is rendered.

maurochip (approved):

peluko00 (approved):
LGTM!

sersanchus (approved):

PR #421, opened 2025-11-10
[18.0][FIX] web_leaflet_lib: use unminified library

NL66278:
Odoo minifies javascript sources when adding those to asset bundles. Having the unminified versions available is needed for debugging. The OCA contributor guidelines state: 'Never add minified Javascript libraries'.


legalsylvain:
That makes sense ! 

Thanks.

Naive question, in which module are you using Leaflib in V18 ? 

/ocabot merge patch

NL66278:
@legalsylvain I am using it in a custom module, that also uses leaflet.draw to plot area's of land used by farmers (but could be used by anybody to draw plots). Intending to offer those modules here as well.

PR #422, opened 2025-11-12
[18.0][ADD] new module web_leaflet_draw_lib

NL66278:


legalsylvain:
Hi @NL66278
- thanks for your contribution.
- how to test this feature ? is it possible to add some demo view to test it on runboat ? 


other question : do you think it's better to create a dedicated module for that "pugin", or better to put it in the main module ? 
- main library : leaflet js : 14k lines
- plugin markercluster included in main module : 2,7 k lines
- plugin leaflet draw : 4k lines.

That's an open question, I don't have a clear point of view.

NL66278:
@legalsylvain It can be tested by also including this PR: https://github.com/OCA/geospatial/pull/429. Maybe I should base 429 on this one? I tested locally. Anyway you can see from the code how it is supposed to be used.

I like to have the leaflet_draw module separate from the plain leaflet one, as the included libraries are also separate, although related projects.

legalsylvain:
> I like to have the leaflet_draw module separate from the plain leaflet one, as the included libraries are also separate, although related projects.

OK Then ! 

/ocabot merge nobump

PR #426, opened 2025-11-18
[18.0][IMP] web_view_leaflet_map: Fixed an error that occurred when no markers were available.

thibaultrey:
Fixed an error that occurs if none of the partners displayed on the map have geographical coordinates; an error occurs because it is not possible to define the map's boundaries and zoom level.

Creation of the leaflet URL parameter without value during installation, even if the demo data is not loaded. This parameter must always be filled in according to the instructions mentioned in the readme file.


bealdav (approved):
Tested on production database, it works all is fine. Thanks

web_view_leaflet_map/static/src/components/map-component/map_view.esm.js
```diff
@@ -161,8 +161,10 @@ export class MapRenderer extends Component {
                 marker.addTo(this.leafletFeatureGroup);
             }
         }
-
-        this.leafletMap.fitBounds(this.leafletFeatureGroup.getBounds().pad(0.1));
+        const bounds = this.leafletFeatureGroup.getBounds();
+        if (bounds.isValid()) {
+            this.leafletMap.fitBounds(bounds.pad(0.1));
+        }
```
legalsylvain, line 167: thanks for this patch !

legalsylvain (commented):
Hi @thibaultrey. Thanks for your PR.

adding a data ir_config_parameter, allways loaded is indeed great.
however, I don't understand the reason to replace the demo ir_config_parameter per a post_init_hook.

I see it less analyzable.

what is the benefit ?

web_view_leaflet_map/static/src/components/map-component/map_view.esm.js
bealdav: @legalsylvain  About hook, if you add parameter in data with noupdate=1 then you can't update by xml in demo data the config parameter And it's required to set noupdate=1

bealdav: And we shouldn't spam map provider demo data for big odoo instance

legalsylvain (commented):
/ocabot merge patch

PR #427, opened 2025-11-18
[18.0][IMP] web_leaflet.* modules. use markercluster.

legalsylvain:
Rational: for the time being, nearby markers are not very visible. With this plugin, markers are grouped if it's not possible to display all markers. also it is possible to see markers that are in the same address. (that is not possible for the time being)

description of the introduced plugin here : https://github.com/Leaflet/Leaflet.markercluster

this plugin has been "recommanded" by the leaflet maintainer team. see https://leafletjs.com/plugins.html#clusteringdecluttering 

Note : I was thinking about an option at the beginning, but in fact I don't see any case where we don't want this plugin, so I don't see the interest to make code more complex.

zoom in : 

image

zoom out : 
image

@mourad-ehm, @bealdav, @thibaultrey : could you take a look on this one ?


bealdav (approved):
Nice it works nicely

thibaultrey:
Hi @legalsylvain 

Thanks for this, I just tested it and it seems to me that we have lost the map's focus on the available markers.
For example, if all your contacts displayed on the map are located in Europe, the focus remains on default coordinate, which gives the illusion that the map is empty.

Could the fitBounds call be kept in order to focus on the right part of the map ?

legalsylvain:
> Could the fitBounds call be kept in order to focus on the right part of the map ?

Hi @thibaultrey. thanks for your review ! 

I didn't know the fitbounds feature. I restored the code and added a comment.

Could you test again ?

thibaultrey (approved):

thibaultrey:
> > Could the fitBounds call be kept in order to focus on the right part of the map ?
> 
> Hi @thibaultrey. thanks for your review !
> 
> I didn't know the fitbounds feature. I restored the code and added a comment.
> 
> Could you test again ?

Thank you @legalsylvain 
Tested and seems to work well 👌🏻

legalsylvain:
Thanks ! 

/ocabot merge minor

PR #429, opened 2025-11-26
[18.][ADD] New module geospatial_plot

NL66278:
New module allows to define and draw plots on a map and link them to partners.

This PR depends on https://github.com/OCA/geospatial/pull/422


legalsylvain:
hi @NL66278 Could you take a look on red CI ? 

thanks !

NL66278:
@legalsylvain The tests is failing because the PR for web_leaflet_draw_lib has not been merged yet. Did you finally get merge rights, or is that also still stalled?

thomaspaulb:
@NL66278 New test failure looks legit: `TypeError: ResPartner._compute_land_acreage() missing 1 required positional argument: 'acreage_sqm'`

legalsylvain:
@NL66278 : what about to have leaflet in the name of the module ? In fact, we could imagine other implementation, based on other librairies. Don't you think ?

NL66278:
> @NL66278 : what about to have leaflet in the name of the module ? In fact, we could imagine other implementation, based on other librairies. Don't you think ?

In this case I would like to keep the name. The use of the leaflet library is very helpful, but still an implementation detail, even though it is unlikely to change in the foreseeable future.

ivs-cetmix (approved):
/ocabot merge patch

PR #435, opened 2026-01-20
[FIX #412] Load openlayers js separately (because it is not minifieble)

### LINKED ISSUE #412: [18.0] OpenLayers assets not loaded in standard mode (Odoo 18) → ol is not defined
# Description:

In Odoo 18, the OpenLayers JS/CSS have been moved into assets (see [base_geoengine/__manifest__.py](https://github.com/OCA/geospatial/blob/18.0/base_geoengine/__manifest__.py?utm_source=chatgpt.com#L32)
).

When creating a custom module in standard mode, these assets are not loaded, and we get the following error in the browser console:
_Uncaught (in promise) ReferenceError: ol is not defined_

This happens because the OpenLayers bundle is not part of the custom module’s assets by default.
However, if we run the instance with ?debug=assets, everything works fine, since all files are included individually.

We can see the same error during the migration of the demo module (cf. [PR #406](https://github.com/OCA/geospatial/pull/406?utm_source=chatgpt.com)
).

# Expected behavior:
OpenLayers should be available automatically to any module that depends on base_geoengine, as in previous versions.

# Possible solutions:

Include the OpenLayers bundle in web.assets_backend directly in base_geoengine, so all dependent modules get it. Or provide clear documentation/example on how to properly include the OpenLayers bundle in custom modules.

Environment:

Odoo version: 18.0
Module: base_geoengine
Browser console shows: ol is not defined

katyukha:
openlayers JS library (ol.js) could not be minified by Odoo minifier, because it is already minified. Thus, load it separately without minification (do not include in bundle)

### DIFF
--- a/base_geoengine/__manifest__.py
+++ b/base_geoengine/__manifest__.py
@@ -30,7 +30,6 @@
             ("include", "web._assets_bootstrap"),
         ],
         "base_geoengine.assets_jsLibs_geoengine": [
-            "/base_geoengine/static/lib/ol-10.5.0/ol.js",
             "/base_geoengine/static/lib/chromajs-3.1.2/chroma.js",
             "/base_geoengine/static/lib/geostats-2.1.0/geostats.js",
             "/base_geoengine/static/lib/geostats-2.1.0/geostats.css",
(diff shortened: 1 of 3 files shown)


katyukha:
@cormaza @legalsylvain @yvaucher @zamberjo Please, review this fix.

Thanks a lot!

flotho (approved):
Hi, 

Thanks.
Make sense, 
Code review,
LGTM

yvaucher (approved):
LGTM

cormaza (approved):
LGTM

PR #441, opened 2026-03-08
[FIX] Added missing `.sudo` or `compute_sudo=True`

katyukha:
This PR adds `_compute_sudo=True` to compute methods that access system models like (ir.model, ir.model.fields, ...) to avoid unexpected access errors in user interface.

Also, this PR adds `.sudo()` when code accesses (reads) system models for same reasone.

### DIFF
--- a/base_geoengine/models/base.py
+++ b/base_geoengine/models/base.py
@@ -69,7 +69,7 @@ class Base(models.AbstractModel):
         field_obj = self.env["ir.model.fields"]
         if not in_tuple:
             return in_tuple
-        name = field_obj.browse(in_tuple[0]).name
+        name = field_obj.sudo().browse(in_tuple[0]).name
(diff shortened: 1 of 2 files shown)


katyukha:
@cormaza @legalsylvain @yvaucher @zamberjo Please, review this fix.

Thanks a lot!

legalsylvain:
Hi. Could yiu explain the bug you want to solve ?

(With step to reproduce)

zamberjo:
I agree with @legalsylvain . We have been using this functionality for a long time without needing these privilege escalations.  We need steps to reproduce the issue and a test case to verify it.

yvaucher:
Seems dubious to me to add some random `sudo` please provide some example of error you encounter.

legalsylvain (changes requested):
see comments.

PR #417, opened 2025-10-08
[16.0][IMP] base_geoengine: Fix Vector Layer related Error

anusriNPS:
Notifying user to define selected attribute_field_id of supported type from vector layer as part of geoengine view xml defintion which avoids below observed JS error.
 
image

PR #442, opened 2026-03-09
[FIX] dot files: updated from template

anusriNPS:
To include https://github.com/OCA/oca-addons-repo-template/pull/333 which resolves
pre-commit issue seen: https://github.com/OCA/oca-addons-repo-template/issues/334

PR #454, opened 2026-04-21
[16.0][IMP] base_geoengine: Avoid error seen clicking on RecordPanel

anusriNPS:
Depends on #417. When a record in record panel does not contain latitude and
longitude details, it fails with error "feature is null". So, avoiding this error and notifying the user to update required details to view the record.

Below is the error seen without latitude and longitude details:

image

### DIFF
--- a/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
+++ b/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
@@ -24,6 +24,7 @@ import {
     reactive,
     useState,
 } from "@odoo/owl";
+import {sprintf} from "@web/core/utils/strings";
 
 /* CONSTANTS */
 const DEFAULT_BEGIN_COLOR = "#FFFFFF";
@@ -53,6 +54,7 @@ export class GeoengineRenderer extends Component {
         this.orm = useService("orm");
         this.view = useService("view");
         this.user = useService("user");
+        this.notification = useService("notification");
 
         // For related model we need to load all the service needed by RelationalModel
         this.services = {};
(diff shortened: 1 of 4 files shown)


anusriNPS:
pre-commit fails which requires #442 to be merged which helps to update template

quirino95 (approved):
Code and functional review: LGTM!

base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
```diff
@@ -571,6 +572,19 @@ export class GeoengineRenderer extends Component {
     onDisplayPopupRecord(record) {
         const popup = this.getPopup();
         const feature = this.vectorSource.getFeatureById(record.resId);
+        if (feature === null) {
+            this.notification.add(
+                this.env._t(
+                    "Please update latitude and longitude details for: " +
+                        record.data.display_name
+                ),
```
HekkiMelody: **chore**: This will not work for translations, only literal strings can be translated

https://www.odoo.com/documentation/18.0/developer/howtos/translations.html

Can you test formatting it like this to see if it's exported properly in the `.pot` file?

https://github.com/OCA/OCB/blob/18.0/addons/base_import/static/src/import_action/import_action.js#L227-L229

base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
```diff
@@ -571,6 +572,19 @@ export class GeoengineRenderer extends Component {
     onDisplayPopupRecord(record) {
         const popup = this.getPopup();
         const feature = this.vectorSource.getFeatureById(record.resId);
+        if (feature === null) {
```
HekkiMelody, line 576: **question**: this `if` can probably replace the other `if` at line 588

for example

```.js
if (feature) {
    ...
    return false;
}
this.mountGeoengineRecord({
...
```

What do you think?

HekkiMelody (changes requested):
Code review. Please specify in the first message that this PR depends on #417

base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
anusriNPS: Updated pot files for newly introduced notification as part of the PR and translations are working as expected

image

base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js
anusriNPS: Agree, the new `if` block is sufficient and hence removed if block which checks for feature is not undefined.

HekkiMelody (approved):
Code review, LGTM

PR #462, opened 2026-05-31
[FIX] web_view_leaflet_map: reload markers on search domain change

Jonasaaaa:
## Problem

On the `leaflet_map` view, applying a search **filter** or **favorite** does
not update the markers. The renderer loads its records only in `onWillStart`
and never reacts to later domain changes; `onPatched` re-renders the markers
but from the same, stale `this.records`. As a result the map keeps showing all
records regardless of the active search.

## Fix

Add an `onWillUpdateProps` hook that reloads the records with the new
`domain`/`context` before the component is patched. `loadRecords` now accepts
an optional `domain`/`context` (falling back to `this.props`), so the filtered
result is fetched and `onPatched` then re-renders the up-to-date markers.

No change to the view API; filtering now works on every `leaflet_map` view.

## How to test

1. Open any action using a `leaflet_map` view with a search view that has
   filters (e.g. `web_view_leaflet_map_partner`).
2. Apply a filter or favorite.
3. Before: markers are unchanged. After: markers reflect the filtered records.

### DIFF
--- a/web_view_leaflet_map/__manifest__.py
+++ b/web_view_leaflet_map/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Leaflet Map View (OpenStreetMap)",
     "summary": "Add new 'leaflet_map' view, to display markers.",
-    "version": "18.0.1.1.2",
+    "version": "18.0.1.1.3",
     "author": "GRAP, Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


Jonasaaaa:
Closing: the fix belongs on the 19.0 line (built on the ongoing migration PR #459), not on 18.0. Superseded accordingly.

PR #464, opened 2026-06-17
[19.0][MIG] web_leaflet_lib: Migration to 19.0

agiuseppe28:
Migration of `web_leaflet_lib` to 19.0.

- History preserved via `git format-patch | git am` from 18.0 (original authors kept).
- Version bumped to `19.0.1.0.0`.
- Depends only on `base`; no functional change — vendored Leaflet assets carried over as-is.
- Module has no tests; installs cleanly on a fresh 19.0 database; the `ir.http.session_info()` override and the `ir.config_parameter` defaults (`leaflet.tile_url`, `leaflet.copyright`) were verified locally, and the static Leaflet assets are served correctly.

`web_leaflet_lib` was not listed in the migration checklist of #413 (added on 18.0 after that issue was generated — see the missing-modules FAQ).

cc @legalsylvain (original author/maintainer).


web_leaflet_lib/readme/CONTRIBUTORS.md
```diff
@@ -1 +1,2 @@
 - Sylvain LE GAL (<https://www.twitter.com/legalsylvain>)
+- Giuseppe Ardoselli (<https://github.com/agiuseppe28>)
```
legalsylvain: thanks for porting this module.
Not sure that asking to an IA to migrate a module, and just bump a version in a manifest is a "contribution".
Don't you think ?

agiuseppe28: No problem I reverted CONTRIBUTORS.md to your entry only. You're right that a version bump isn't much of a contribution and to be fair credit wasn't even the point.  I need the module on 19.0 for some downstream work and would rather upstream the migration than carry a fork, also because the module was already migration ready, maybe someone could benefit from this. If PR still needs something else, I'd be glad to address it.

legalsylvain (approved):
diff review. no test.

thanks !

PR #474, opened 2026-08-06
[18.0][MIG] website_geoengine: Migration to 18.0

LINTOANTONY007:
This PR is depended on https://github.com/OCA/partner-contact/pull/2417
Once this PR is merged ,

1) website_geoengine will be migrated to 18.0
2) Field `opening_hours` in `res.partner` model is moved to `partner-contact/partner_store` module,  as `opening_hours` is a store related property field. So I think it must be in module `partner_store` of `partner-contact `repo.

Please Refer : https://github.com/OCA/partner-contact/pull/2417

3) Added some Test Cases.


LINTOANTONY007:
@mymage , Can you please review this PR ?

LINTOANTONY007:
@tarteo , Can you review this PR?

website_geoengine/controllers/res_partner.py
```diff
@@ -13,7 +13,7 @@ def tags(self, **kw):
     def partners(self, **kw):
         tags = kw.get("tags", {})
         lang = kw.get("lang", "en_US")
-        maxResults = kw.get("maxResults", "200")
+        max_results = kw.get("maxResults", "200")
```
tarteo: Should `maxResults` key also be snake case, or is this somehow required?

website_geoengine/controllers/__init__.py
```diff
@@ -1 +1,2 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
tarteo: Nitpicking: IMO adding license header to \_\_init\_\_.py file is silly

website_geoengine/controllers/res_partner.py
LINTOANTONY007: if we need to make `maxResults`  key  snake case , then I need to make change in  other module (Refer the line mentioned below). It is doable , simply because of this I just neglected it. If that is the case I can do it. 

https://github.com/OCA/geospatial/blob/4c2b269624e1324736c0634e5742265707d7c5f0/website_geoengine_store_locator/static/src/js/snippets/s_openlayer_store_locator/search.esm.js#L309

```
 const args = {
            tags: tags,
            lang: this.lang,
            maxResults: this.maxResults,
        };
```

website_geoengine/controllers/__init__.py
LINTOANTONY007: Removed it !

website_geoengine/controllers/res_partner.py
LINTOANTONY007: @tarteo , Shall I do it ?

LINTOANTONY007: @tarteo , Done ! , changed key to snake case.

PR #479, opened 2026-09-08
[IMP] base_geoengine: Work around missing required field

SirPyTech:
**Steps**:

1. In the filesystem, delete a view from module `database_cleanup`
2. Install `base_geoengine` and `database_cleanup`
3. Restore the deleted view
4. Update `database_cleanup`

**Current**:
ERROR: null value in column "projection" of relation "ir_ui_view" violates not-null constraint

**Expected**:
The module updates

Work around for https://github.com/odoo/odoo/issues/286530.


HekkiMelody (approved):
LGTM.

This is necessary to avoid a sneaky but scary update error.

PR #481, opened 2026-09-16
[17.0][FIX] base_geoengine: Fix Vector Layer related Error

anusriNPS:
FW #417


max3903 (commented):
## Code review

**Verdict: approve with nits** — sensible forward-port of merged #417; please fix the `it.po` header regression before merge.

### What it does
When a vector layer’s `attribute_field_id` isn’t available on the loaded records (typically missing from the geoengine view fields), styling/legend used to blow up in JS. This adds `checkAttributeFieldUsage()`, shows a warning notification, and skips style/legend setup.

### Good
- Matches the already-merged 16.0 fix (#417); same approach.
- Early-outs around `styleVectorLayer` / `initLegend` / `setStyle` avoid cascading errors.
- User-facing `_t` + notification is better than a console exception.
- CI is green.

### Should fix
1. **`it.po` header wipe** — this diff drops `Language: it`, Weblate `Plural-Forms`, `Last-Translator`, and `X-Generator`. Please only add the new msgid and restore the previous header metadata.
2. **Null-safe `attribute_field_id`** — `extractLayerValues` still does `cfg.attribute_field_id[1]` with no guard. Prefer an early return in `checkAttributeFieldUsage`:

```js
if (!cfg.attribute_field_id) {
    return true;
}
```

(Same concern was already raised on #417.)

### Nits
- Warning can fire on every domain/layer refresh → possible notification spam.
- PR body is only `FW #417` — a short “why” would help reviewers.
- Please add the `needs review` label.

### Process
Age 0 days / 0 approvals so far — not merge-ready yet under OCA policy (5 days + 2 reviews). Related consumer: OCA/field-service#1621.

